PR #4, opened 2019-06-20
Move non-EHR code to other modules

bbimber:
The GeneticsCore module was originally just ONPRC-EHR related; however, over time evolved as a place to for me to put code for sequence-related ONPRC applications that are not EHR, but not appropriate for our more general modules.  This PR is a two step process that initially removes a lot of code from GeneticsCore module, leaving it with just the EHR-specific parts.  

There will be a separate commit to trunk with two new modules where this code was moved.  There wll also be a separate PR to the distribution project.


bbimber:
@labkey-jeckels I apparently cannot assign reviewers, but this is what we talked about.

PR #28, opened 2020-10-06
Make a minimal dev build easier via Artifactory

labkey-jeckels:
#### Rationale
We want devs to be building just the modules they're editing, but that requires improved dependency declarations

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/70

#### Changes
* Explicitly declare that we need more whole modules so they get pulled from Artifactory if they're not available locally.

* Don't double-declare dependency in modules.properties and build.gradle

### DIFF
--- a/onprc_ehr/build.gradle
+++ b/onprc_ehr/build.gradle
@@ -11,5 +11,8 @@ dependencies
    modules("org.labkey.module:laboratory:${labkeyVersion}@module") { transitive = true }
    modules("org.labkey.module:LDK:${labkeyVersion}@module") { transitive = true }
    modules("org.labkey.module:dataintegration:${labkeyVersion}@module") { transitive = true }
-   modules("org.labkey.module:premium:${labkeyVersion}@module") { transitive = true }
+BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: ":server:modules:LabDevKitModules:laboratory", depProjectConfig: 'published', depExtension: 'module')
+BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: ":server:modules:LabDevKitModules:LDK", depProjectConfig: 'published', depExtension: 'module')
+BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: ":server:modules:dataintegration", depProjectConfig: 'published', depExtension: 'module')
+BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: ":server:modules:premium", depProjectConfig: 'published', depExtension: 'module')
 }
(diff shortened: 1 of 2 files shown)


onprc_ehr/build.gradle
```diff
@@ -5,4 +5,11 @@ dependencies
    implementation "com.sun.mail:jakarta.mail:${javaMailVersion}"
    BuildUtils.addLabKeyDependency(project: project, config: "implementation", depProjectPath: ":server:modules:LabDevKitModules:LDK", depProjectConfig: "apiJarFile")
    BuildUtils.addLabKeyDependency(project: project, config: "implementation", depProjectPath: ":server:modules:ehrModules:ehr", depProjectConfig: "apiJarFile")
+
+   // Ensure that we have these modules, either from Artifactory or local source, even if they aren't direct dependencies
+   // to make it easier to build a minimal set on a dev machine
+   modules("org.labkey.module:laboratory:${labkeyVersion}@module") { transitive = true }
+   modules("org.labkey.module:LDK:${labkeyVersion}@module") { transitive = true }
+   modules("org.labkey.module:dataintegration:${labkeyVersion}@module") { transitive = true }
+   modules("org.labkey.module:premium:${labkeyVersion}@module") { transitive = true }
```
labkey-susanh: I know I directed you this way, but I think we should switch this to use `BuildUtils.addLabKeyDependency` instead as that will work when you don't have an enlistment and will build the modules if you do have an enlistment.
```suggestion
BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: ":server:modules:LabDevKitModules:laboratory", depProjectConfig: 'published', depExtension: 'module')
BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: ":server:modules:LabDevKitModules:LDK", depProjectConfig: 'published', depExtension: 'module')
BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: ":server:modules:dataintegration", depProjectConfig: 'published', depExtension: 'module')
BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: ":server:modules:premium", depProjectConfig: 'published', depExtension: 'module')
```

labkey-susanh (changes requested):

labkey-jeckels:
@labkey-susanh I tested with both permutations and it built the local version of LDK and Laboratory if I had the source checked out and pulled from Artifactory if I didn't. I don't know exactly how to test without an enlistment, but going with your updated suggestion

labkey-susanh (approved):
It should be the case that if you don't have an enlistment in, say, LabDevKitModules, or it isn't included as a gradle project in your settings.gradle file, the dependency declaration will essentially be the same as what you had before and it will pull from Artifactory.  If you DO have an enlistment and it is includes as a project, you can still turn off building by setting buildFromSource property to false for a particular module.

PR #55, opened 2021-01-05
Merge onprc19.1Prod changes from revision 65972-66122

labkey-jeckels:


extscheduler/resources/schemas/dbscripts/sqlserver/extscheduler-20.936-20.937.sql
```diff
@@ -0,0 +1,13 @@
...
+CREATE TABLE [extscheduler].[TempCoV19Final](
+    [container] [uniqueidentifier] NOT NULL,
+    [UserName] [nvarchar](4000) NULL,
+    [UserID] [int] NULL,
+    [EventID] [int] NULL,
+    [StartDate] [smalldatetime] NULL,
+    [EmployeeID] [nvarchar](4000) NULL,
+    [Created] [smalldatetime] NULL
+    ) ON [PRIMARY]
+    GO
```
labkey-bpatel, line 13: Do we also need to update the schema version to 20.937 in ExtSchedulerModule? currently it is at 20.603.

onprc_ehr/src/org/labkey/onprc_ehr/notification/ColonyAlertsNotification.java
```diff
@@ -1061,7 +1099,7 @@ protected void pmicSchedulerAlert(final Container c, User u, final StringBuilder
             return;
         }
         //Daily events query
-        TableInfo ti = QueryService.get().getUserSchema(u, c, "extscheduler").getTable("PMIC_Scheduler");
+        TableInfo ti = QueryService.get().getUserSchema(u, c, "extscheduler").getTable("PMIC_Scheduler_Daily");
```
labkey-bpatel, line 1102: I didn't see PMIC_Scheduler_Daily created anywhere in this merge, & further investigating looks like it is not there in the module either, so perhaps stored in the db. Makes sense for it to be part of the module, unless its outside the scope of this merge.

labkey-bpatel:
There are a couple of test failures with pipeline errors during study load, previously/in my experience, they have helped troubleshoot troublesome/broken queries.

onprc_ehr/src/org/labkey/onprc_ehr/notification/ColonyAlertsNotification.java
labkey-jeckels: I'm not finding it on the production server either. This should be addressed, but it won't be causing test failures so I'd like to proceed with the merge. I can ask them about it.

labkey-jeckels: I'm not finding it on the production server either. This should be addressed, but it won't be causing test failures so I'd like to proceed with the merge. I can ask them about it.

extscheduler/resources/schemas/dbscripts/sqlserver/extscheduler-20.936-20.937.sql
labkey-jeckels: Yes, this file had a merge conflict which I've now resolved and bumped the version number.

labkey-jeckels:
Fixes pushed for query validation. New test run should come back happy.

labkey-bpatel (approved):

PR #58, opened 2021-01-11
20.11 fb ext scheduler update

jonesgaohsu:
#### Rationale
The existing merge attempted to create tables that already existed and the sql string should have included a drop if exists statement

#### Related Pull Requests
Issue was found with last merge relating to extscheduler files

#### Changes
Changed each script to insure it included a drop if exists statement


labkey-jeckels:
I changed the target branch to release20.11-SNAPSHOT instead of develop.

@jonesgaohsu are the EXEC core.fn_dropifexists calls not working here. They did in my local testing. I switched to use that approach instead of DROP IF EXISTS because TeamCity is runs tests on SQL Server 2012, and DROP IF EXISTS wasn't introduced until 2016.

labkey-jeckels:
D

labkey-jeckels:
Determined to be unneeded after a subsequent deployment succeeded

PR #68, opened 2021-01-25
Ticket 42312: Missing submenu selection from Research Procedure entry from

labkey-jeckels:
#### Rationale
Restore the missing "Copy previous task" menu item, which didn't get fully merged to the 20.11 branch

#### Changes
* Add dependency on ONPRC's CopyTaskWindow.js


labkey-bpatel (approved):

jonesgaohsu:
Tested and works as expected

PR #72, opened 2021-01-29
Fix SQL Server 2012 testing on TeamCity - remove redundant DROP IF EXISTS

labkey-jeckels:
#### Rationale
TeamCity runs tests on SQL Server 2012 since we officially support it, and this script is using a syntax that's not supported
 
#### Changes
* Remove redundant DROP IF EXISTS, as previous statement has already used the 2012-compatible stored procedure to drop it


labkey-tchad (approved):
Verifying on SQL Server 2012: https://teamcity.labkey.org/buildConfiguration/LabKey_2011Release_Internal_InternalSuites_OldDependenciesSqlserver/1245350

labkey-jeckels:
After followup commit, server has started cleanly for test run so proceeding with merge.

PR #77, opened 2021-02-04
20.11 fb mease45 day alert fix

jonesgaohsu:
#### Rationale
Report was causing errors, LK updated the query and it now produces values in Dev

#### Related Pull Requests
* 

#### Changes
SQL Query was updated and testing in Dev Worked as desired

### DIFF
--- a/ONPRC_EHR_ComplianceDB/resources/queries/EHR_ComplianceDB/Measles45DayAlert.sql
+++ b/ONPRC_EHR_ComplianceDB/resources/queries/EHR_ComplianceDB/Measles45DayAlert.sql
@@ -1,47 +1,49 @@
+--update kollil 2021-02-04
+
 SELECT
-	emp.employeeid,
-	emp.Name,
-	emp.email,
-	emp.unit,
-	emp.category,
-	emp.location,
-	emp.majorudds,
-	emp.notes,
-	emp.requirementname,
-	--c.date,
-	Case
-       When (c.requirementname is null) Then emp.date
-       When (c.requirementname is not null) Then c.date
-       ELSE ''
-	End AS DateCompleted,
-	emp.comment,
-	emp.trainer,
-	Case
-       When (c.requirementname is null) Then TIMESTAMPDIFF('SQL_TSI_DAY', emp.date, now())
-       When (c.requirementname is not null) Then '-1'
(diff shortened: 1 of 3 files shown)


ONPRC_EHR_ComplianceDB/resources/queries/EHR_ComplianceDB/Measles45DayAlert.sql
```diff
@@ -1,47 +1,49 @@
...
+        When (c.requirementname is null) Then emp.date
+        When (c.requirementname is not null) Then c.date
+        ELSE ''
+        End AS DateCompleted,
+    emp.comment,
+    emp.trainer,
+    Case
+        When (c.requirementname is null) Then CAST(TIMESTAMPDIFF('SQL_TSI_DAY', emp.date, now()) AS VARCHAR)
+        When (c.requirementname is not null) Then CAST('-1' AS VARCHAR)
+        ELSE ''
```
labkey-jeckels: Do you really want this to be VARCHAR? It won't sort or filter the way that users might want. Also, the ELSE will never happen because requirementname will either be null or not null.

```suggestion
        When (c.requirementname is null) Then TIMESTAMPDIFF('SQL_TSI_DAY', emp.date, now())
        When (c.requirementname is not null) Then -1
```

labkey-jeckels (changes requested):

jonesgaohsu:
So what is the correct way to update a feature branch, just correct and push?


Gary Jones
Business Analyst 3
Oregon National Primate Research Center
Beaverton Oregon 97006

Office 503-346-5556
Cell 971-998-7909

labkey-jeckels (approved):

labkey-jeckels:
@jonesgaohsu Your followup push of the changes looks fine. In general, you can push as many changes as you like to a feature branch, so that's the right process.

Before creating the feature branch, you weren't up to date with the latest from the release20.11-SNAPSHOT. That's the fourth step here:

https://www.labkey.org/ONPRC/wiki-page.view?name=DevProcess

I fixed it up this time and we can cover how you can do this yourself in the future.

jonesgaohsu:
THis update has been verified in our test instance and is ready to update

PR #86, opened 2021-02-13
Modified the program to prevent medication formulary default data fro…

Ohsudev:
…m reloading when medication code was not the control that was modified.

#### Changes
Modified the medication entry screen so that changes to medication code selection drop down listings will be the only condition to reload the default values from  the formulary.


labkey-jeckels (approved):
See my comments in the corresponding ticket:

https://www.labkey.org/ONPRC/Support%20Tickets/issues-details.view?issueId=42479

PR #92, opened 2021-02-20
Workaround for TreatmentToDrug ETL ObjectId value in the target

labkey-jeckels:
#### Rationale
TreatmentToDrug ETL is failing because it's pulling Treatment Order table's ObjectId value and inserting it into the Drug table. That causes a duplicate key violation when there multiple doses for a single day

#### Changes
* Use SQL expression so that the association with ObjectId is lost
* Rename files to match casing that's already present in ONPRC DB, preventing an error creating the experiment run that's describing the ETL


jonesgaohsu (approved):
Corrections approved and in review on testing environment

jonesgaohsu:
Pushed to Default

PR #104, opened 2021-03-04
20.11_ misc_charges_links

jonesgaohsu:
#### Rationale
Update of Xml for Misc Charges to add links to invoiced Items and Items Invoiced

#### Related Pull Requests
* 

#### Changes
Misccharges.query.xml


labkey-jeckels:
Gary, the feature branch has a typo in its name. It should be "_fb_" instead of "_gn_".

I recommend recreating the feature branch from scratch with the intended change. This should address both the feature branch naming and the extra edit to the SQL query.

PR #108, opened 2021-03-10
Add initial MHC ETL

bbimber:
@jonesgaohsu @labkey-jeckels We discussed migrating all MHC data and associated pipelines from prime to prime-seq, and to make an ETL that will pull aggregated MHC results from prime-seq back to prime. This is the latter. I dont know if I'm targeting the desired branch here, but my hope is that we could merge this into whatever branch PRIMe is running and then deploy to the test server. I need a little more time to get the source data ready, but I think it's worth getting this ETL prepared so that this part is good to go.

PR #111, opened 2021-03-10
Create ETLs for MHC transfer from prime-seq to prime

josheckels:
From @bbimber: We discussed migrating all MHC data and associated pipelines from prime to prime-seq, and to make an ETL that will pull aggregated MHC results from prime-seq back to prime. This is the latter. I dont know if I'm targeting the desired branch here, but my hope is that we could merge this into whatever branch PRIMe is running and then deploy to the test server. I need a little more time to get the source data ready, but I think it's worth getting this ETL prepared so that this part is good to go.

Equivalent to https://github.com/LabKey/onprcEHRModules/pull/108 with tweaked FB name and target branch

### DIFF
--- a/GeneticsCore/resources/queries/geneticscore/mhc_data_source.sql
+++ b/GeneticsCore/resources/queries/geneticscore/mhc_data_source.sql
@@ -1,0 +1,41 @@
+SELECT
+  s.subjectId as Id,
+  s.allele,
+  --s.primerPairs,
+  s.shortName,
+  s.totalRecords as testsPerformed,
+  s.status as result,
+  cast('SSP' as varchar) as type
+
+FROM assay.SSP_assay.SSP.SSP_Summary s
+WHERE s.subjectId is not null
+
+UNION ALL
+
+SELECT
+  a.subjectId as Id,
+  a.marker as allele,
+  null as shortName,
+  count(*) as totalTests,
+  cast('POS' as varchar) as result,
+  cast('SBT' as varchar) as type
+
+FROM assay.GenotypeAssay.Genotype.Data a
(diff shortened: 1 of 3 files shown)


josheckels:
@bbimber I invited you to be a committer to this repo to make it easier if are additional changes are needed.

GeneticsCore/resources/queries/geneticscore/mhc_data_source.sql
```diff
@@ -0,0 +1,41 @@
+SELECT
+  s.subjectId as Id,
+  s.allele,
+  --s.primerPairs,
+  s.shortName,
+  s.totalRecords as testsPerformed,
```
labkey-ankurj:
```suggestion
  s.totalRecords as totalTests,
```

labkey-ankurj (changes requested):
added a suggestion.

GeneticsCore/resources/queries/geneticscore/mhc_data_source.sql
josheckels: @bbimber This looks like the right change to me to get things aligned between the SQL and XML, but can you confirm?

bbimber:
@josheckels @labkey-ankurj yeah, that's a reasonable change. it seems to be a long-standing issue from another query (not causing issues), but should be updated. 

However, I dont see to have commit permission here. is it possible to update that?

josheckels:
Committed. You should have an invite waiting to get write access to this repro:

https://github.com/LabKey/onprcEHRModules/invitations

labkey-ankurj (approved):

PR #114, opened 2021-03-12
Disallow null Id values in onprc_ehr.AvailableBloodVolume, give it a PK

josheckels:
#### Rationale
While adding automated test coverage for the ONPRC blood volume validation code, we discovered that having duplicate rows for a given animal in onprc_ehr.AvailableBloodVolume will cause errors when doing the validation. It also lacks a PK, making it hard for the tests to interact with it to populate testing data

#### Changes
* Make the Id column NOT NULL and identify it as the primary key

### DIFF
--- a/extscheduler/module.properties
+++ b/extscheduler/module.properties
@@ -1,6 +1,5 @@
 Name: extscheduler
 ModuleClass: org.labkey.extscheduler.ExtSchedulerModule
-ModuleDependencies: LDK
 SupportedDatabases: mssql
 RequiredServerVersion: 15.2
 ManageVersion: false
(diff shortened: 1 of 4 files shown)


josheckels:
@jonesgaohsuand @kollil can you validate this change for us? I think it will help prevent problems and lets us add the test coverage.

There's a chance it would interfere with how the data is being populated. If, for example, it loads all new rows and then deletes all of the old rows, it would violate the PK constraint, and the sequence would need to be reversed (delete the existing rows first, then load the replacement data)

labkey-sweta (approved):

PR #132, opened 2021-04-01
20.11 fb pmic updates

kollil:
#### Rationale
PMIC Scheduler queries are updated and removed two fields from the data entry panels.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/extscheduler/resources/queries/extscheduler/PMIC_Scheduler_Daily.sql
+++ b/extscheduler/resources/queries/extscheduler/PMIC_Scheduler_Daily.sql
@@ -19,5 +19,5 @@ SELECT
     e.modified
 FROM Events e, Resources r
 Where r.id = e.resourceid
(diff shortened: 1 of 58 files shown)


labkey-jeckels:
@kollil I think that you hadn't pulled the latest from release20.11-SNAPSHOT when you created this branch. Hence, it's showing all of the files from your original PMIC feature branch. Please recreate the branch. You can use this URL to get a patch that contains what I think are the only edits you intended to make:

https://github.com/LabKey/onprcEHRModules/commit/ae45611f62b0570c66c04c5246fecf83d8bc7cdf.diff

It looks like you re-introduced the ContainerIDs in the WHERE clauses. As I said in my earlier review, we don't want to embed the EntityIDs like that as it will make it impossible to test on LabKey dev machines, etc. You may be able to query based on a container name or other criteria to get the same results.

PR #134, opened 2021-04-01
20.11 fb mhc

jonesgaohsu:
#### Rationale
Copyuing the MHC type etl to the default folder for ETLs in onprc_EHR

#### Related Pull Requests
* 

#### Changes

MHC_Typing.xml


jonesgaohsu:
I had forgotten to change the branch, this has been resolved so it should go thru

labkey-jeckels:
@jonesgaohsu This branch appears to be based on the already-merged 20.11_fb_mhc branch. Can you create a new branch with just the changes you want to make?

If you want to move the ETL from GeneticsCore to ONPRC_EHR, that seems reasonable to me, but I think we should delete the original location so we don't end up with two ETLs with the same name. @bbimber - heads up on the proposed change.

bbimber:
I dont get this. what branch is deployed to prime? it is not release-20.11? if it's not release-20.11, we should just merge the original PR as-is to that branch. i am happy to do that.

i do not believe the moving of this ETL was on purpose, or I dont think there's any reason we necessarily need to.

labkey-jeckels:
The original PR has already been merged to release20.11-SNAPSHOT, which is the branch that's being deployed to PRIMe production.

https://github.com/LabKey/onprcEHRModules/pull/111

I see a commit from Gary in this branch from yesterday that copied the ETL's XML file:

https://github.com/LabKey/onprcEHRModules/pull/134/commits

Gary, should we discuss this on Tuesday's call?

PR #139, opened 2021-04-07
20.11 fb behavior alert

kollil:
#### Rationale
A new "notes" alert is added to the Behavior notifications.

### DIFF
--- a/onprc_ehr/src/org/labkey/onprc_ehr/notification/BehaviorNotification.java
+++ b/onprc_ehr/src/org/labkey/onprc_ehr/notification/BehaviorNotification.java
@@ -146,8 +146,11 @@ public class BehaviorNotification extends ColonyAlertsNotification
     }
 
     /*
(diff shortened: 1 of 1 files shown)


kollil:
Josh, I added only one file BehaviorNotification.java file to this new fb. Don't know how the previous commits got added to this pull. Please remove them.

Thanks
lakshmi

labkey-jeckels (approved):
The reason so many files were showing up was because you hadn't set the target branch to release20.11-SNAPSHOT. It was still the default, which is "develop", which doesn't yet have all of the other changes you've merged into release20.11-SNAPSHOT.

Anytime you have unexpected files showing in the list in the pull request it's a good idea to double-check the source and target branches.

kollil:
Yes, Josh, you are right. I got confused as it showed me "Compare & Pull Request" button instead of just "Pull request" button. I did something to edit that as I didn't have anything to compare. 
I will be careful next time.

PR #142, opened 2021-04-12
Add PMIC and IPC datasets to study archive to fix query validation

labkey-jeckels:
#### Rationale
New SQL queries were added recently that depend on newly added study datasets. We need those in the study archive to keep the tests happy during query validation.

#### Changes
* Add missing new PMIC and IPC datasets
* Reference shared column configuration in organ_weights


labkey-jeckels:
@labkey-martyp @labkey-bpatel this is how I discovered the repro for https://github.com/LabKey/ehrModules/pull/168

labkey-bpatel (approved):

PR #144, opened 2021-04-15
20.11 fb covid s cheduler temp tables

jonesgaohsu:
#### Rationale
This is a series of SQL Scripts that update the public temp files used for Covid 19 Processing of Scheduler items

#### Related Pull Requests
* 

#### Changes
Added script 401 to 409 and updated the extSchedulerModule number to reflect the change

### DIFF
--- a/onprc_ehr/resources/data/procedure_drug_templates.tsv
+++ b/onprc_ehr/resources/data/procedure_drug_templates.tsv
@@ -1,4 +1,4 @@
-value	sort_order
+`value	sort_order
 None
 Analgesia - Standard
 Analgesia - Biopsies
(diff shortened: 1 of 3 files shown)


onprc_ehr/resources/data/procedure_drug_templates.tsv
```diff
@@ -1,4 +1,4 @@
-value	sort_order
+`value	sort_order
```
labkey-jeckels, line 1: This looks like a typo, correct?

labkey-jeckels (changes requested):

jonesgaohsu:
Newer request with updated files being loaded

PR #146, opened 2021-04-21
20.11 fb surgery procedure update

Ohsudev:
#### Changes
Added the Procedure category into Procedure selection drop down listings and have this feature applied to both Surgery input screens.


Ohsudev:
Can you please exclude the two files above that seems to think there is a conflict.  I didn't include those two files into my feature branch.

Thanks,
Raymond

labkey-jeckels (approved):

PR #168, opened 2021-05-25
Created a new column to track training categories that are not requir…

Ohsudev:
…ed as training requirements.

#### Changes
Created a new column to track training categories that are not required as training requirements.  This set is for the onprcEHRModule version. A separate feature branch was created for EHRModule version


labkey-jeckels (approved):

Ohsudev:
No longer needed.

PR #175, opened 2021-06-03
Modified setting to allow "units" field to be a selectable field when…

Ohsudev:
… using the Bulk Edit window.

#### Changes
Modified settings to allow "units" field to be a selectable field when using the Bulk Edit window.


labkey-jeckels:
@kollil A quick reminder after our last call - the plan is that I'll do the final review on these changes but that someone at ONPRC will do an initial review and approval.

kollil (approved):

labkey-jeckels (approved):
Looks good.

There are test failures but they have already been fixed in the release20.11-SNAPSHOT branch so we can ignore them in this feature branch.

kollil:
Josh, I approved them, sorry for the delay, it totally slipped from my mind. I was out of office this morning and started my day late.

PR #181, opened 2021-06-15
Support pipeline job to aggregate MHC data prior to ETL

bbimber:


bbimber:
This is related to ONPRC issue: https://prime.ohsu.edu/issues/ONPRC/Admin/IS/ISE%20Issues%20and%20Training/Issues/details.view?issueId=7323

GeneticsCore/src/org/labkey/GeneticsCore/GeneticsCoreController.java
```diff
@@ -18,4 +32,73 @@ public GeneticsCoreController()
...
+                    MhcAggregationPipelineJob.saveLastRun(getContainer(), null);
+                }
+
+                PipeRoot pipelineRoot = PipelineService.get().findPipelineRoot(getContainer());
+                MhcAggregationPipelineJob job = new MhcAggregationPipelineJob(getContainer(), getUser(), getViewContext().getActionURL(), pipelineRoot);
+                PipelineService.get().queueJob(job);
+            }
+            catch (Exception e)
+            {
+                _log.error(e);
```
labkey-jeckels: This will log the toString() of the exception, not the full stack trace. Add a message argument if you want the full stack.

labkey-jeckels (commented):
I'll create a LK side FB and PR to get TeamCity to build

labkey-jeckels:
@bbimber before I do that, PRIMe is on 20.11 right now. I don't know your timetable for wanting to deploy this. Is 21.3 the desired branch?

GeneticsCore/src/org/labkey/GeneticsCore/GeneticsCoreController.java
bbimber: > @bbimber before I do that, PRIMe is on 20.11 right now. I don't know your timetable for wanting to deploy this. Is 21.3 the desired branch?

I see - I thought it was newer. That's a tough call. So I only need the one ETL XML file deployed to PRIMe. Most of this, including the pipeline job, runs on prime-seq, which is 21.3. Perhaps I could split this into two PRs that separates the ETL file?

bbimber: On logging: for some reason I thought what I was doing worked. One question, is this safe? I think e.getMessage() is nullable, right? Do I need a null check or would it be better practice to just add my own message?

_log.error(e.getMessage(), e);

vs.

_log.error("Error starting MHC pipeline", e);

labkey-jeckels: I prefer the second approach, where you give a message so that the log has a little more context, but it should be safe to pass null without creating an NPE

labkey-jeckels:
@bbimber Yes, creating a second PR for 20.11 is probably the easiest then.

labkey-jeckels:
I created the LK-side FB. I had to correct the FB name so I opened a new PR:

https://github.com/LabKey/onprcEHRModules/pull/188

PR #187, opened 2021-06-17
Add ETL to sync MHC data from prime-seq to prime

bbimber:
@labkey-jeckels  This is related to #181. It separates just the ETL into a release20.11-SNAPSHOT PR. This is the minimal file I need deployed to PRIMe for this ETL. The other PR can be 21.3 and will be deployed to prime-seq.

I think we want release20.11-SNAPSHOT, not release20.11, right?


labkey-jeckels:
Created LK FB with corrected naming convention and putting the file in /resources/etls/ instead of /resources/etl/

https://github.com/LabKey/onprcEHRModules/pull/189

PR #193, opened 2021-06-19
Backport schema changes for MHC ETL

bbimber:
@labkey-jeckels If you remember, I am splitting the MHC ETL changes into those targeting PRIMe/20.11, and those targeting prime-seq/213. I missed one change we need on PRIMe. This backports the changes in the mhc_data table to 20.11 so we can deploy to prime. Note: I fixed the PG "set" error you caught on develop here. This might make merge conflicts downstream, but i'm not sure how to best approach this otherwise

### DIFF
--- a/GeneticsCore/resources/etl/MHC_Typing.xml
+++ b/GeneticsCore/resources/etl/MHC_Typing.xml
@@ -5,7 +5,7 @@
     <transforms>
         <transform type="RemoteQueryTransformStep" id="assay">
             <description>Copy to target</description>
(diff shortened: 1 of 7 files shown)


labkey-jeckels:
These changes look OK to me but I will defer to Trey on the possibilities of merge conflicts. I suspect if there are any, it'll be trivial to resolve them, but better to have an expert weigh in.

labkey-tchad (approved):
Not really any way to avoid conflicts if they're going to happen. Backports are usually pretty easy to resolve though.

PR #201, opened 2021-06-24
Delete ONPRC_EHR/MHC_Typing.xml

bbimber:
I'm not sure the origin of this file, but my intent was to put the MHC_Typing.xml file in GeneticsCore module, not ONPRC_EHR (see #189). I might have done this wrong at some point (there was a lot of manual moving around to put files into 21.3 and 20.11). Anyway, this should go in favor of the MHC_Typing.xml already merged into GeneticsCore module.


labkey-jeckels (approved):
This looks like a safe enough change that we can skip the TC build/test, but here's the reference for FB naming to get it to match with that TC is looking at:

https://www.labkey.org/Documentation/wiki-page.view?name=git_branch_naming

You don't need the "-SNAPSHOT-" in the branch name. The PR should always be against the SNAPSHOT branch though.

bbimber:
At some point I will figure this branch naming out...

PR #219, opened 2021-07-23
Update EHR MHC query to point to new aggregated data

bbimber:
This is hopefully the final commit for the MHC aggregation / ETL project


labkey-jeckels (approved):
Thanks for making these FBs directly in the LK onprcEHRModules repo. It eliminates a number of steps in the process.

PR #237, opened 2021-09-09
Update MHC_Data_Animals.sql

bbimber:
This change is related to the prior MHC-data changes. The data of reference for MHC now resides on prime-seq. As a result, we need to base this query (which governs other queries to show which animals have MHC data), on the cached MHC data table


labkey-jeckels (changes requested):
PRIMe production is on 20.11, with a planned upgrade to 21.7 later this month. I'm guessing you want to target one of those branches instead of 20.7?

PR #278, opened 2021-10-29
Fix ONPRC test query validation errors 

labkey-jeckels:
#### Rationale
Fix tests broken by query validation failures

#### Changes
* Add missing column to flags dataset
* Enable geneticscore module
* Use method to hook in special handling


labkey-sweta (approved):

labkey-jeckels:
@jonesgaohsu note that this PR deletes RandData_MHCData.sql which had a number of errors that preventing it from running successfully (referencing the wrong table name, for example). This removes it to get the tests passing. We can recreate it when it's ready for full usage.

PR #286, opened 2021-11-09
Correct field name in MHC query

bbimber:
I missed this with my earlier MHC commit


labkey-jeckels (approved):

bbimber:
thanks @labkey-jeckels

PR #290, opened 2021-11-15
Don't try to run ONPRC_EHR_ComplianceDB when on Postgres

labkey-jeckels:
#### Rationale
We're seeing intermittent failures in the crawler like:

https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_Premium_Ehr_DailyEhrPostgres/1605326?buildTab=overview

#### Changes
* Mark ONPRC_EHR_ComplianceDB as SQL Server only


labkey-jeckels:
We'll take this via this 21.7 branch instead:

https://github.com/LabKey/onprcEHRModules/pull/255

PR #293, opened 2021-11-17
Use `BuildUtils.addLabKeyDependency` for declaring the module dependency

labkey-susanh:
#### Rationale
The dependency declaration to get the SampleManagement module deployed that doesn't use `addLabKeyDependency` causes problems when there is no version of the SampleManagement module already published (such as when we are first building our release versions to publish).  This change will pick up the locally built module if present and otherwise get the module from Artifactory.

One question here is whether it is a problem that this dependency declaration will cause any server distribution containing this module to include the SM module.  This is probably only a question for our test suites.

#### Changes
* Update method for declaring server module dependency.


labkey-tchad (approved):

labkey-jeckels:
@labkey-susanh this won't make SM a dependency of ONPRC_EHR in terms of what's enabled in a given folder, correct?

labkey-susanh:
> @labkey-susanh this won't make SM a dependency of ONPRC_EHR in terms of what's enabled in a given folder, correct?

@labkey-jeckels Correct.  Same dependency declaration as before, just more flexible.

labkey-jeckels (approved):
Sounds good to me then.

PR #296, opened 2021-11-18
Fix crawler failure - Declare explicit dependency on EHR JS files

labkey-jeckels:
#### Rationale
https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_Premium_Ehr_DailyEhrSqlserver/1610374

#### Changes
* Declare dependency on EHR JS files instead of relying on them to be pulled in via other modules being enabled in the container

### DIFF
--- a/ONPRC_EHR_ComplianceDB/resources/views/begin.view.xml
+++ b/ONPRC_EHR_ComplianceDB/resources/views/begin.view.xml
@@ -1,5 +1,7 @@
 <view xmlns="http://labkey.org/data/xml/view" title="Compliance and Training">
     <dependencies>
         <dependency path="LDK/LDKApi"/>
+        <dependency path="ehr.context" />
+        <dependency path="ehr/ehr_api"/>
     </dependencies>
 </view>


labkey-jeckels:
@labkey-tchad @labkey-sweta I got this to fully repro for me locally. On my dev machine, the client side JS error is reported in the server log files like this:

ERROR AdminController          2021-11-18T10:34:24,705 tps-jsse-nio-8443-exec-7 : Client exception detected:
https://localhost:8443/labkey/panorama/onprc_ehr_compliancedb-begin.view?
https://localhost:8443/labkey/panorama/onprc_ehr_compliancedb-begin.view?
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.69 Safari/537.36
jeckels@labkey.com
Uncaught ReferenceError: EHR is not defined
  https://localhost:8443/labkey/panorama/onprc_ehr_compliancedb-begin.view?:565:18
  https://localhost:8443/labkey/ext-4.2.1/ext-all-sandbox-debug.js:9869:28
  constructor.fire (https://localhost:8443/labkey/ext-4.2.1/ext-all-sandbox-debug.js:9824:49)
  constructor.Ext.apply.readyEvent.readyEvent.fire (https://localhost:8443/labkey/ext-4.2.1/ext-all-sandbox-debug.js:10058:52)
  fireReadyEvent (https://localhost:8443/labkey/ext-4.2.1/ext-all-sandbox-debug.js:10136:28)
  https://localhost:8443/labkey/ext-4.2.1/ext-all-sandbox-debug.js:2094:27


The crawler failure indicates that the config on TeamCity expect this to be captured as well. Is there something we can do to make it more generally available? It would have made troubleshooting this much faster.

labkey-tchad:
We disable the JS error logging while the crawler is running. Historically, it would just log numerous false-positives; the rapid navigation of the crawler tended to interrupt ajax calls which would often end up causing JavaScript exceptions.
That being said, that change was made in 2013 so maybe its worth revisiting.

labkey-jeckels:
@labkey-tchad I think we made a somewhat centralized change to stop treating cancelled AJAX requests (due to page navigation) as error conditions, but it's been a long time so I couldn't point to a specific commit. Good if this is only turned off for the crawler, but could be a reasonable experiment to try a daily run with it turned on

labkey-adam (approved):

PR #305, opened 2021-11-26
Added location the two alerts: Flags and Notes

kollil:
Added location field to Notes and Flags alerts


Ohsudev (approved):

labkey-jeckels (approved):

labkey-jeckels:
This change caused the tests on TeamCity to fail because the test setup didn't include the category field on the study.flags dataset. I added it here:

https://github.com/LabKey/onprcEHRModules/pull/305/commits/d2b94e66ee68c01b45a5a94f48eb418ef2b0ec9f

@keith-labkey let's discuss this update in a future call so that it can be part of the ONPRC team's test review process.

PR #307, opened 2021-12-01
Created a new report that will sort Active Clinical case report using…

Ohsudev:
#### Changes
Created a new report that will sort Active Clinical case report using the cage location that resides within the P2 values.


kollil (approved):
looks good

labkey-jeckels (approved):

PR #337, opened 2021-12-23
Don't assume JS Date object has a format() function

labkey-jeckels:
#### Rationale
Testing the ASB form in Safari, I was getting a JS error about not having a format() function on the Date object

#### Changes
* Switch to use the Ext4 function which we know will be available


Ohsudev (approved):

labkey-jeckels:
I branched from 21.11 instead of 21.7. New PR coming momentarily.

PR #354, opened 2022-01-17
Drop unused stored procedure that was created in "dbo" instead of "onprc_ehr"

labkey-adam:
#### Rationale
This procedure is unused and the fact that it was created in `dbo` causes problems when attempting to re-install the `onprc_ehr` module


labkey-jeckels (approved):
Change looks fine. Should we also cut it from the script that creates it?

labkey-adam:
We could, but I hesitate to do more than the minimum here since we're not consolidating or managing this schema. There are 55 DROP statements throughout the `onprc_ehr` scripts... if we coalesce one shouldn't we coalesce them all? Are there other unused procedures that should be dropped?

labkey-jeckels:
That's fine. I was thinking we might hit permission problems on some DBs trying to make changes in DBO, but I guess it hasn't been a problem so why rock the boat now.

PR #403, opened 2022-04-21
Modified version of Compliance Editor panel section.

Ohsudev:
#### Changes
Modified Compliance Editor screen to only use Labkey panel code instead of using wiki panel.


Ohsudev:
Merged feature branch 21.7 into 22.3 Compliance Editor permission panel.

PR #413, opened 2022-04-29
Modified Printable reports for diet medication online report to displ…

Ohsudev:
#### Changes
Modified online version of the diet reports so that the report display diet medication information from Prime history reports instead of the Clinical snapshot.


kollil (approved):
Looks good

PR #421, opened 2022-05-04
Delete incorrectly named ehrSLA.xml linked schema template file

labkey-jeckels:
#### Rationale
https://github.com/LabKey/onprcEHRModules/pull/418 included changes that broke all tests in the ONPRC suite and is bleeding out into other test suites that happen to run with the ONPRC modules deployed.

It looks like ehrSLA.xml was an attempted to define a new linked schema template, but it is missing the .template.xml file extension. I assume that ehrSLA.template.xml is a subsequent, successful attempt at the template.

#### Changes
* Delete the ehrSLA.xml file that is causing failures because there's no ehrSLA schema in the database


kollil:
@labkey-jeckels, Yes, it was a wrong file I created without the .template.xml extension and corrected it later. I have safely deleted the file from intellij. The ehrsla.xml file can be deleted.

Thanks
Lakshmi

kollil (approved):
ehrSLA.xml file is invalid file.

PR #423, opened 2022-05-06
22.3 fb necropsy request form update

Ohsudev:
>

#### Changes
Merge changes made from 21.7 into 22.3 for the Necropsy request form.


kollil (approved):
Looks good

onprc_ehr/resources/web/onprc_ehr/model/sources/NecropsyRequest.js
```diff
@@ -0,0 +1,180 @@
...
+                defaultValue: 'DCM: Pathology Services'
+            },
+            chargeId: {
+                allowBlank: false,
+                nullable: false,
+                lookup: {
+                    filterArray: [
+                        LABKEY.Filter.create('active', true, LABKEY.Filter.Types.EQUAL),
+                        LABKEY.Filter.create('category', 'Pathology', LABKEY.Filter.Types.EQUAL),
+                        LABKEY.Filter.create('rowid', '4484;4485;4486;4487;4488;4489;5296;5297;5298;4491;4492', LABKEY.Filter.Types.EQUALS_ONE_OF)
```
labkey-martyp: Hi Raymond @Ohsudev, can you filter on something here besides rowid?  That is an auto increment value that you really have no control over.  This only fits the very specific case that is currently on your database.  It won't work in any other environment and could easily be broken.  We also can't enter test data to match this filter.  Can you instead filter on itemCode and name starts with Pathology?

PR #429, opened 2022-05-10
Modified Compliance Reports to include training exemptions

Ohsudev:
#### Changes
Modified Compliance Reports to include training exemptions.


kollil (approved):
Looks good

PR #449, opened 2022-05-30
SLA Breeding data

kollil:
Updated the URL for the SLA breeding report to point to the linked schema , ehrSLA


kollil:
Closing this PR as ehrSLA.xml file showed up again even though I didn't add the file.

PR #452, opened 2022-05-31
SLA breeding report fix

kollil:
Updated the URL for the breeding report so it picks the data from the linked schema, ehrSLA


kollil:
@keith-labkey , It has been a challenge for me creating any fb. The ehrSLA file gets created unknowingly. I created one yesterday and then deleted it. I created a separate fb (22.3_fb_Delete_ehrSLA) to delete the ehrSLA.xml file. Just created another fb and it got created again. :-(

PR #467, opened 2022-07-07
Modified to allow formatting of the weight's values three decimal pla…

Ohsudev:
#### Changes

Modified to allow formatting of the weight's values three decimal places.


kollil (approved):
looks good

PR #469, opened 2022-07-12
Modified Printable reports by Vets so that we can include room locati…

Ohsudev:
#### Changes
Modified Printable reports to allow Active Clinical case to include an option to generate report by
room locations and by vet name.


kollil (approved):
Looks good

PR #470, opened 2022-07-14
Modified to include dependency files for Epoc Input form

Ohsudev:
#### Changes
Modified to include dependency files for Epoc Input form


kollil (approved):
looks good

PR #480, opened 2022-07-30
Modified Birth form to allow a special permission group to be able to…

Ohsudev:
>

#### Changes
Modified Birth form to allow a special permission group to be able to release the Birth form screen other than Sit admins.


kollil (approved):
Looks good

PR #490, opened 2022-09-09
Modified program to display a short list of investigators with active…

Ohsudev:
#### Changes
Modified program to display a short list of investigators with active assignments.


kollil (approved):
Looks good

PR #493, opened 2022-09-15
Modified Active Clinical case by Vet to allow all room locations to g…

Ohsudev:
#### Changes
Modified Active Clinical case by Vet to allow all room locations to generate a report.


kollil (approved):
Looks good

PR #507, opened 2022-10-03
Modified Vet_assignment data set to show only active protocol items.

Ohsudev:
Modified Vet_assignment data set to show only active protocol items.


kollil (approved):
looks good

PR #520, opened 2022-10-12
22.3 fb request form approved notification

Ohsudev:
Modified program to implement a process to generate email notification when a request is approved.


onprc_ehr/resources/scripts/onprc_ehr/onprc_triggers.js
```diff
@@ -1165,7 +1165,110 @@ exports.init = function(EHR){
...
+    EHR.Server.TriggerManager.registerHandlerForQuery(EHR.Server.TriggerManager.Events.BEFORE_UPSERT, 'ehr', 'requests', function (helper, scriptErrors, row, oldRow) {
+        if (row.requestid && row.qcstate.label ==  'Request: Approved'){
+
+                var requestid = [];
+                for (var i in row.requestid.length){
+                    var rows = row.requestid[i];
+                    requestid.push(i);
+                }
+
+        }
```
labkey-martyp: You don't need this

onprc_ehr/resources/scripts/onprc_ehr/onprc_triggers.js
```diff
@@ -1165,7 +1165,110 @@ exports.init = function(EHR){
...
+
+            var msgs =   helper.getJavaHelper().sendRequestStateEmail("Request: Approved", row.requestid );
+            if (msgs && msgs.length){
+                LABKEY.ExtAdapter.each(msgs, function(msg){
+                    EHR.Server.Utils.addError(scriptErrors, 'qcstate', msg, 'INFO');
+                }, this);
+            }
+
+        }
+    });
```
labkey-martyp: You have two of these AFTER_UPSERT triggers for requests. Looks like this one might be the duplicate that can be deleted.

onprc_ehr/resources/scripts/onprc_ehr/onprc_triggers.js
```diff
@@ -1165,7 +1165,110 @@ exports.init = function(EHR){
...
+                    requestid.push(i);
+                }
+
+        }
     });
 
+    //Added: 10-4-2022  R.Blasa
+    EHR.Server.TriggerManager.registerHandlerForQuery(EHR.Server.TriggerManager.Events.AFTER_UPSERT, 'ehr', 'requests', function(helper, errors, row, oldRow){
+        if (row.requestid && row.qcstate.label ==  'Request: Approved'){
+            var requestsApproved = triggerHelper.getApprovedArray();
```
labkey-martyp: I don't see getApprovedArray function in this PR. But I don't think you need to do any extra handling here for batch updates. row.requestid should always be a single value.

onprc_ehr/resources/scripts/onprc_ehr/onprc_triggers.js
```diff
@@ -1165,7 +1165,110 @@ exports.init = function(EHR){
...
+            var msgs =   helper.getJavaHelper().sendRequestStateEmail("Request: Approved", requestsApproved );
```
labkey-martyp:
```suggestion
            var msgs =   helper.getJavaHelper().sendRequestStateEmail("Request: Approved", [row.requestid] );
```
row.requestid should always be a single value in this handler so just making a single value array out of it should work fine.

labkey-martyp (commented):
Hi Raymond, I added some comments. Since it doesn't look like you'll be needing to support batch updates, I think you can simplify the trigger. I did not test this PR but added some suggestions that I think are all you need for this.

labkey-martyp:
Ok Raymond, I added the approved request tracking to the EHR as it may be helpful for others and made the TriggerScriptHelper.sendRequestStateEmail public for use here. That [PR is here](https://github.com/LabKey/ehrModules/pull/424), I will try to get that merged today.

I'll put a patch in the ticket (it won't let me attach it here).  Once the EHR PR is merged in, you can apply the patch to this fb and it will use that new EHR functionality and send notifications (batched if necessary) on request approval.

kollil (approved):

PR #523, opened 2022-10-15
Modified program to add the remark values into the Urinalysis result …

Ohsudev:
Modified program to add the remark values into the Urinalysis result Full History reports.


kollil (approved):

onprc_ehr/src/org/labkey/onprc_ehr/history/ONPRCUrinalysisLabworkType.java
```diff
@@ -18,17 +18,14 @@
...
- */
+//Modified: 10-14-2022 R.Blasa
 public class ONPRCUrinalysisLabworkType extends SortingLabworkType
 {
     public ONPRCUrinalysisLabworkType(Module module)
     {
         super("Urinalysis", "study", "Urinalysis Results", "Urinalysis", module);
         _resultField = null;
         _qualResultField = "results";
+        _testIDfieldremark = "remark";
```
labkey-martyp: You shouldn't need this, there's already a _remarkField you can use instead of _testIDfieldremark

onprc_ehr/src/org/labkey/onprc_ehr/history/ONPRCUrinalysisLabworkType.java
```diff
@@ -18,17 +18,14 @@
...
     }
```
labkey-martyp, line 34: You'll probably just want to override the getLine function here with your changes.

```
    protected String getLine(Results rs, boolean redacted) throws SQLException
    {
        StringBuilder sb = new StringBuilder();
        String testId = getTestId(rs);
        Double result = _resultField == null ?  null : rs.getDouble(FieldKey.fromString(_resultField));
        if (rs.wasNull())
        {
            result = null;
        }
        String units = _unitsField == null ?  null : rs.getString(FieldKey.fromString(_unitsField));
        String qualResult = _qualResultField == null ?  null : rs.getString(FieldKey.fromString(_qualResultField));
        String testIDfieldremark = _remarkField == null ?  null : rs.getString(FieldKey.fromString(_remarkField));

        if (result != null || qualResult != null || testIDfieldremark != null )
        {
            sb.append("<td style='padding: 2px;'>").append(testId).append(": ").append("</td>");
            sb.append("<td style='padding: 2px;'>");

            boolean unitsAppended = false;
            boolean remarkAppended = false;
            if (result != null)
            {
                sb.append(result);
                if (units != null)
                {
                    sb.append(" ").append(units);
                    unitsAppended = true;
                }
                if (testIDfieldremark != null)
                {
                    sb.append(" ").append(testIDfieldremark);
                }
            }

            if (qualResult != null)
            {
                if (result != null)
                    sb.append(" ").append(result);

                sb.append(qualResult);

                if (units != null && !unitsAppended)
                    sb.append(" ").append(units);

            }

            if (testIDfieldremark != null)
            {
                if (result != null)
                    sb.append(result);

                if (qualResult!= null)
                    sb.append(" ");

                if (units != null && !unitsAppended)
                    sb.append(" ");

                if (!remarkAppended)
                    sb.append(" ").append(testIDfieldremark);
            }




            sb.append("</td>");

            //append normals
            String normalRange = _normalRangeField == null ?  null : rs.getString(FieldKey.fromString(_normalRangeField));
            String status = _normalRangeStatusField == null ?  null : rs.getString(FieldKey.fromString(_normalRangeStatusField));
            if (normalRange != null)
            {
                if (status != null)
                {
                    String color = "green";
                    if (status.equals("High"))
                        color = "#E3170D";
                    else if (status.equals("Low"))
                        color = "#FBEC5D";

                    sb.append("<td style='padding: 2px;background-color: " + color + ";'>&nbsp;" + status + "&nbsp;</td>");
                }

                sb.append("<td style='padding: 2px;'>");
                sb.append(" (").append(normalRange).append(")");
                sb.append("</td>");
            }
        }

        return sb.toString();
    }`

labkey-martyp (commented):
Instead of updating the ehr module, can you try this change below?

onprc_ehr/src/org/labkey/onprc_ehr/history/ONPRCUrinalysisLabworkType.java
Ohsudev: I created _testIDFieldremark.  You meant that I should use a_remarkField.  Is that correct?

labkey-martyp: Yes, both are just string variables for the word "remark" and they never change.  It's not holding any data value, just telling what column name to look at.

If you prefer to use a second variable for "remark" (_testIDfieldremark) that's fine it won't break anything logically but it should be defined here in this file, not in ehr. The code I provided in the next comment uses _remarkField, you will need to replace it there if using _testIDfieldremark.

PR #575, opened 2023-01-04
Update kinship and inbreeding coefficient tests

labkey-martyp:
#### Rationale
Kinship calculations have been updated. Update the genetic tests here and add more thorough testing of kinship and inbreeding results.

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/493

#### Changes
* Add specific pedigree scenarios to the test data
* Verify kinship coefficients and inbreeding coefficents of those scenarios

### DIFF
--- a/onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
+++ b/onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
@@ -67,7 +67,6 @@ import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -690,7 +689,7 @@ public class ONPRC_EHRTest extends AbstractGenericONPRC_EHRTest
         goToProjectHome();
         beginAtAnimalHistoryTab();
 
-        String id = ID_PREFIX + 1;
+        String id = ID_PREFIX + 10;
         AnimalHistoryPage animalHistoryPage = new AnimalHistoryPage(getDriver());
 
         animalHistoryPage.searchSingleAnimal(id);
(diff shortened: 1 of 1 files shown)


onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
```diff
@@ -1258,20 +1276,42 @@ public void testGeneticsPipeline() throws Exception
...
         animalHistoryPage.refreshReport();
         animalHistoryPage.clickCategoryTab("Genetics")
                 .clickReportTab("Kinship");
 
+        log("verify kinship");
         assertTrue(isElementPresent(Locator.linkWithText("CLICK HERE TO LIMIT TO ANIMALS IN SELECTION")));
 
         DataRegionTable kinshipTable = animalHistoryPage.getActiveReportDataRegion();
-        assertEquals("Incorrect number of rows before limiting animal selection", 3, kinshipTable.getDataRowCount());
+        assertEquals("Incorrect number of kinship rows before limiting animal selection", 48, kinshipTable.getDataRowCount()); // TODO: should be more
```
labkey-jeckels: Is this TODO relevant?

labkey-jeckels (changes requested):
The update test is failing on a duplicate row problem during the inserts.

labkey-tchad (commented):
The test failure is because `createBirthRecords` no longer calls `deleteIfExists` before inserting birth records.
If you make `_hasCreatedBirthRecords` static, `createBirthRecords` should no-op for the second test to call it.
Alternatively, you could have `doSetup` create the birth records instead of making the individual tests do it. I can't imagine it takes terribly long, since its all done via API.

labkey-jeckels (approved):

onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest.java
labkey-martyp: Removed

PR #618, opened 2023-02-15
Modified program to fix an issue with a software bug that prevented t…

Ohsudev:
Modified program to fix an issue with a software bug that prevented the time portion of the date from being recorded.


kollil (approved):
Raymond, Is the

PR #666, opened 2023-04-13
Add test for procedures for section

labkey-martyp:
#### Rationale
A recent change caused the remarks section not to be auto populated when selecting a procedure in the procedures (encounters) form section. This tests that use case (now that it's fixed).

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/576

#### Changes
* Add ONPRC_EHRTest2.testProceduresFormSection

### DIFF
--- a/onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest2.java
+++ b/onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest2.java
@@ -529,6 +529,30 @@ public class ONPRC_EHRTest2 extends AbstractONPRC_EHRTest
     }
 
     @Test
+    public void testProceduresFormSection() throws Exception
+    {
+        _helper.goToTaskForm("Research Procedures", "Submit Final", false);
+
+        log("testing add row");
+        Ext4GridRef grid = _helper.getExt4GridForFormSection("Procedures");
+        grid.clickTbarButton("Add");
+        grid.waitForRowCount(1);
+        grid.completeEdit();
+
+        log("testing procedure remark");
+        Ext4FieldRef procedure = grid.getActiveEditor(1, "procedureid");
+        procedure.clickTrigger();
+        click(Locator.tagContainingText("li", "Achilles Tendon Repair"));
+        grid.completeEdit();
+
+        final Object remark = grid.getFieldValue(1, "remark");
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest2.java
```diff
@@ -528,6 +528,30 @@ public void testFlagsApi() throws Exception
...
+
+        log("testing procedure remark");
+        Ext4FieldRef procedure = grid.getActiveEditor(1, "procedureid");
+        procedure.clickTrigger();
+        click(Locator.tagContainingText("li", "Achilles Tendon Repair"));
+        grid.completeEdit();
+
+        final Object remark = grid.getFieldValue(1, "remark");
+        waitFor(() -> remark != null && remark.toString().equals("Intubated with ___ ID cuffed ET tube, maintained on 1.5% isoflurane and 1 L/min oxygen. Prepared skin with PVI, lavaged with sterile saline. Prepared skin with PVI, aseptically draped. Incised over the caudolateral surface of the tendon, 5 cm. Identified the gastronemius, superficial digital flexor, and common tendon. Freshened tendon edges and apposed with 2-0 PDS with a far-near near-far pattern. Lavaged, closed fascia with 3-0 Monocryl SI. Freshened skin edges, closed with 3-0 Monocryl SI. Placed a padded bandage and fiberglas cast. Administered analgesics. Good recovery from anesthesia."),
+                "Expected Remark not set", WAIT_FOR_JAVASCRIPT);
```
labkey-tchad, line 550: @labkey-martyp This wait doesn't do anything. `remark` isn't going to change. TeamCity hasn't had any problems but I presume the wait was added for a reason.

labkey-martyp: Ok I assumed it was needed because selecting the procedure in the previous lines causes an API query to auto populate the remark. Seemed like it needed a wait in there somewhere and this seemed to be working. I can look into removing that in the next PR I do in this repo.

labkey-tchad: Yeah, it might need a wait. Just looking at the Java code; if the value of `remark` isn't correct, the `waitFor` isn't going to see if the value on the page changes.

PR #675, opened 2023-04-19
Support for restricted issue lists

labkey-klum:
#### Rationale
To support a client request for restricted issue lists we will introduce changes to enable the configuration on a per issue list basis. Restricted issue lists will have the ability to limit access to individual issues at the point at which a user tries to access an issue to:
- view details
- update
- resolve, close, reopen

The feature will not filter access from the issue list view nor will it prevent access of any tables in the query schema. 

Access checking, and storage of admin UI settings is delegated to a new interface `RestrictedIssueProvider` which will need to be registered by the module providing the implementation.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1177
- https://github.com/LabKey/platform/pull/4329

#### Changes
- implement and register a `RestrictedIssueProvider` instance.

### DIFF
--- a/onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
+++ b/onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
@@ -1,0 +1,44 @@
+package org.labkey.onprc_ehr.issues;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.data.Container;
+import org.labkey.api.issues.Issue;
+import org.labkey.api.issues.RestrictedIssueProvider;
+import org.labkey.api.security.Group;
+import org.labkey.api.security.User;
+import org.labkey.api.security.ValidEmail;
+import org.labkey.api.util.Pair;
+
+import java.util.List;
+import java.util.Objects;
+
+public class RestrictedIssueProviderImpl implements RestrictedIssueProvider
+{
+    @Override
+    public boolean hasPermission(Container c, User user, @NotNull Issue issue, @Nullable Group groupWithAccess)
+    {
+        // Site admins always have access
+        if (user.isInSiteAdminGroup())
+            return true;
+
+        // Assigned to users have access
+        if (Objects.equals(issue.getAssignedTo(), user.getUserId()))
+            return true;
+
+        // anyone on the notify list
+        List<Pair<User, ValidEmail>> notifyUsers = issue.getNotifyListUserEmail();
+        for (Pair<User, ValidEmail> nu : notifyUsers)
+        {
+            if (user.equals(nu.getKey()))
+                return true;
+        }
+
+        // if there is an option group configured, members will have access
+        if (groupWithAccess != null)
+        {
+            return user.isInGroup(groupWithAccess.getUserId());
+        }
+        return false;
+    }
+}
--- a/onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
+++ b/onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
@@ -81,7 +81,7 @@ public class RestrictedIssueProviderImpl implements RestrictedIssueProvider
     public boolean hasPermission(User user, @NotNull Issue issue, List<Issue> relatedIssues, List<ValidationError> errors)
     {
         // Site admins always have access
-        if (user.isInSiteAdminGroup())
+        if (user.hasSiteAdminPermission())
             return true;
 
         Container issueContainer = ContainerManager.getForId(issue.getContainerId());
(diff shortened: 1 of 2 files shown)


onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return Boolean.parseBoolean(value);
+        }
+        return false;
+    }
+
+    @Override
+    public void setRestrictedIssueListGroup(Container c, String issueDefName, @Nullable Group group)
+    {
+        setPropertyValue(c, issueDefName, PROP_RESTRICTED_ISSUE_LIST_GROUP, null != group ? String.valueOf(group.getUserId()) : "0");
```
cnathe, line 47: the "0" here is for the null case right? Is there a reason that we couldn't just use `null` for the property value?

onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
```diff
@@ -0,0 +1,136 @@
...
+            Group group = getRestrictedIssueListGroup(issueContainer, issue.getIssueDefName());
+            if (!checkAccess(user, issue, group))
+            {
+                errors.add(new SimpleValidationError("This issue is in a restricted issue list. You do not have access to this issue"));
+                return false;
+            }
+        }
+
+        // the user must also have access to all related issues
+        for (Issue related : relatedIssues)
```
cnathe, line 99: I saw in the platform PR that before calling this `hasPermission` method, it gathers up the `relatedIssues` and passes them through to here. Thoughts on having this method be the one that gets the related issues instead? Is there a scenario where this method is called with a different set of related issues then the ones from the IssueManager call?

onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
```diff
@@ -0,0 +1,136 @@
...
+        if (issueDefName == null)
+            throw new IllegalArgumentException("Issue def name must be specified");
+
+        return CAT_RESTRICTED_ISSUE_DEF_PROPERTIES + issueDefName;
+    }
+
+    @Override
+    public boolean hasPermission(User user, @NotNull Issue issue, List<Issue> relatedIssues, List<ValidationError> errors)
+    {
+        Container issueContainer = ContainerManager.getForId(issue.getContainerId());
```
cnathe, line 87: if the `issueContainer` is null, it looks like we'll fall through to the bottom and return `true`. is that expected? I assume we would probably hit issues before this method if that `issueContainer` was null.

onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+        return true;
+    }
+
+    private boolean checkAccess(User user, @NotNull Issue issue, @Nullable Group groupWithAccess)
+    {
+        // Site admins always have access
+        if (user.isInSiteAdminGroup())
+            return true;
```
cnathe: could move this up to the start of the `hasPermission` method and short circuit all of the issue specific checks if the user is a site admin.

cnathe (approved):

onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
labkey-klum: yes, this becomes a `noop` and most likely will fail in whatever action the issue was requested for.

onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
labkey-klum: Yes that would have looked cleaner but there wasn't a great way to maintain a single choke point with `IssueManager.getIssue` in a way to do access checking in one place. `IssueManager.getIssue` calls the helper method `_getIssue` to get the fully materialized issue and I didn't want to expose that as a public method to allow callers to get an issue without checking or add a flag to suppress access checking.

onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
labkey-klum: sure, that makes sense

onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
```diff
@@ -80,6 +80,10 @@ private String getPropMapName(String issueDefName)
     @Override
     public boolean hasPermission(User user, @NotNull Issue issue, List<Issue> relatedIssues, List<ValidationError> errors)
     {
+        // Site admins always have access
+        if (user.isInSiteAdminGroup())
```
labkey-adam: I believe that `user.hasSiteAdminPermission()` is the preferred check

labkey-adam (approved):
Should switch out the `isInSiteAdminGroup()` call, but no need for re-review

onprc_ehr/src/org/labkey/onprc_ehr/issues/RestrictedIssueProviderImpl.java
labkey-klum: great, thanks!

PR #685, opened 2023-04-28
Modified Demographics PE to include Assigned Vets

Ohsudev:
Modified Demographics PE to include Assigned Vets

### DIFF
--- a/onprc_ehr/resources/queries/study/demographicsPE.sql
+++ b/onprc_ehr/resources/queries/study/demographicsPE.sql
@@ -28,7 +28,8 @@ CASE
   ELSE 'Clinically Restricted'
 END as isRestricted,
        g.date as p2date,
-       g.p2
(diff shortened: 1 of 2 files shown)


onprc_ehr/resources/queries/study/demographicsPE.sql
```diff
@@ -28,7 +28,8 @@ CASE
   ELSE 'Clinically Restricted'
 END as isRestricted,
        g.date as p2date,
-       g.p2
+       g.p2,
+       d.id.assignedVet.assignedVet
```
bbimber, line 32: There's no benefit from adding this here. You are updating the default view below (and not actually referencing this column). Editing the qview.xml file like below will accomplish what you seem to be trying to do here.

bbimber (changes requested):

PR #701, opened 2023-05-16
Strict SQL Fragment Update

labkey-martyp:
#### Rationale
Related PRs enforce stricter SQL fragment syntax. The current implementation of a couple notification SQL queries bring in SQL file comments in the SQL Fragment creation which fails some asserts. Update these to append the SQL files after fragment creation.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4416
* https://github.com/LabKey/platform/pull/4394

#### Changes
* Update SQL fragment creation
* Update some string builder appends and null checks

### DIFF
--- a/onprc_ehr/src/org/labkey/onprc_ehr/notification/ColonyAlertsNotification.java
+++ b/onprc_ehr/src/org/labkey/onprc_ehr/notification/ColonyAlertsNotification.java
@@ -1596,7 +1596,6 @@ public class ColonyAlertsNotification extends AbstractEHRNotification
         SQLFragment sql = ti.getFromSQL("hc");
         Map<String, Object> params = new HashMap<>();
         params.put("MINDATE", cal.getTime());
-        QueryService.get().bindNamedParameters(sql, params);
         sql = new SQLFragment("SELECT * FROM ").append(sql);
         QueryService.get().bindNamedParameters(sql, params);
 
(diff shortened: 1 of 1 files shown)


onprc_ehr/src/org/labkey/onprc_ehr/notification/ColonyAlertsNotification.java
```diff
@@ -1591,27 +1590,29 @@ protected void nonContiguousHousing(final Container c, User u, final StringBuild
...
+        if (null == ti)
+            return;
 
         SQLFragment sql = ti.getFromSQL("hc");
         Map<String, Object> params = new HashMap<>();
         params.put("MINDATE", cal.getTime());
         QueryService.get().bindNamedParameters(sql, params);
-        sql = new SQLFragment("SELECT * FROM " + sql.getSQL(), sql.getParams());
+        sql = new SQLFragment("SELECT * FROM ").append(sql);
         QueryService.get().bindNamedParameters(sql, params);
```
labkey-matthewb, line 1600: Probably harmless, but I don't think we need both calls to bindNamedParameters()

labkey-matthewb: Yup, the second one will definitely be a noop.  Either one will do the job, I'd probably delete the first one.

labkey-matthewb (approved):

PR #742, opened 2023-07-25
Modified program to allow reporting of Animal Group from a provided w…

Ohsudev:
Modified program to allow reporting of Animal Group from a provided web link.


jonesgaohsu (approved):
Code looks solid

PR #767, opened 2023-08-17
PMIC history queries fix

kollil:
Fixed the PMIC procedures history data queries.


jonesgaohsu (approved):
No issues found with the proposed changes

PR #782, opened 2023-09-04
Add action to allow external server to trigger re-import of genetics data

bbimber:
This is related to https://github.com/LabKey/ehrModules/pull/650, and described in more detail there.


jonesgaohsu (approved):
No Issues observed

PR #796, opened 2023-09-19
Removed a section from group processing page

kollil:
#Tkt 9767 -  "Blood needed for viral surveillance" section is removed from the animal group processing page.


jonesgaohsu (approved):
Code looks clean

PR #815, opened 2023-10-10
Vet assignment query edits

kollil:
Update vet assignment queries.


brentlogan (approved):
Approved for production build.

As a subsequent build, we will do a cleanup, deleting unnecessary queries, etc.

PR #823, opened 2023-10-18
Modified Completion Date data set to include two new reporting columns.

Ohsudev:
Modified Completion Date data set to include two new reporting columns.
Host and Employeed End date.


kollil (approved):
Looks good

PR #842, opened 2023-11-07
Backport changes to increase size of geneticscore column

bbimber:
This is backporting #837 and #836.  I didnt think we would need these on PRIMe, but forgot we ETL MHC data to PRIMe already.

### DIFF
--- a/GeneticsCore/resources/schemas/dbscripts/postgresql/geneticscore-17.13-17.14.sql
+++ b/GeneticsCore/resources/schemas/dbscripts/postgresql/geneticscore-17.13-17.14.sql
@@ -1,0 +1,1 @@
+ALTER TABLE geneticscore.mhc_data ALTER COLUMN marker TYPE varchar(1000);
--- a/GeneticsCore/resources/schemas/dbscripts/sqlserver/geneticscore-17.13-17.14.sql
+++ b/GeneticsCore/resources/schemas/dbscripts/sqlserver/geneticscore-17.13-17.14.sql
@@ -1,0 +1,1 @@
+ALTER TABLE geneticscore.mhc_data ALTER COLUMN marker nvarchar(1000);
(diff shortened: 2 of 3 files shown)


bbimber:
Be aware that these are cherry picked from 23.7. I dont think we want to squash this at merge time, in order to preserve the commits and avoid conflicts when 23.3 is eventually merged forward.

jonesgaohsu (approved):
Changes are as expected

bbimber:
@labkey-tchad or @labkey-martyp: this PR is backporting two commits from 23.7 to 23.3 (via git cherry pick). If I squash this PR, does that create any conflicts/issues when this squashed commit is subsequently merged forward back into to 23.7? I do not have the option enabled to do anything other than squash/merge in this repo. Thanks.

bbimber:
@labkey-jeckels: do you by chance have thoughts on my question in the last post? Again, this is a backport of two cherry-picked commits from 23.7. This is approved and I could go ahead a squash/merge, but I thought when backporting we wanted to preserve the commit history (e.g., not squash), to avoid conflicts when it gets merged forward. If the latter is true, could you or someone with admin privs on this repo please do the commit? If I'm wrong about this I can just go ahead and squash myself. Just trying to avoid the potential for issues in the future merges...

labkey-jeckels:
@bbimber I think it would be fine to squash merge this PR, but @labkey-tchad is better versed than I am in the nuances here.

FWIW, I believe that anyone with write permission to the repo can merge without squashing via the Git CLI. We've set up the GitHub PR UI to not offer the option as it's typically not the right merge for our workflow, but it doesn't lock down the real backend permissions. Again @labkey-tchad can confirm/correct.

bbimber:
@labkey-jeckels, ok, thanks for the reply. Just wanted to avoid hassles for merges. I'll wait until EOD in case Trey or anyone wants to weigh in, but if not I will squash/merge this. 

Interesting on the UI vs. CLI...would never have guessed that.

labkey-tchad:
@bbimber There would be conflicts either way. As long as the changes identical, they should be pretty easy to resolve.

bbimber:
OK, thanks. I really thought git was smart about seeing that the commits were identical (i.e. the whole reason not to squash), but I guess not.

labkey-tchad:
You're right about it being smart but I don't think it matters whether you squash the cherry-picked commits or not. I just did a merge locally and there weren't any conflicts.

PR #863, opened 2023-12-04
Replace direct user of char() as string with SqlFragment in GroupConcats

bbimber:
Probably in 23.7, LK introduced changes in how SQL is handled for group_concat(), which is used in a lot of EHR's calculated columns. The important change is that java code adding a group_concat column needs to supply the delimiter as a SQL Fragment, not a string. This patch (which is targeting 23.11, but you can do with as you like), addresses all the example I saw that needed to be changed.

If you run anything >=23.7 as-is, I expect that any of the columns being changed in this PR would show up as:

```
Value1charValue2charValue3
```

Rather than:
```
Value1
Value2
Value3
```

Note: this patch is dependent on a change in LDK module I made in 23.11, so backporting this to 23.7 by itself would not work.

### DIFF
--- a/onprc_ehr/src/org/labkey/onprc_ehr/table/ONPRC_EHRCustomizer.java
+++ b/onprc_ehr/src/org/labkey/onprc_ehr/table/ONPRC_EHRCustomizer.java
@@ -889,12 +889,12 @@ public class ONPRC_EHRCustomizer extends AbstractTableCustomizer
             if (realTable == null)
                 return;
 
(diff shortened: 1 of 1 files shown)


bbimber:
and note, the TeamCity build as-is will fail since my PR is expecting this commit from LDK module: https://github.com/bimberlabinternal/LabDevKitModules/commit/d0dfe3c667b755411cb880a305f07bd2d6747446. That could be easily merged to release-23.11 if you plan to try to fix this in 23.11.

jryurkanin:
@labkey-martyp  can you take a look at this?

labkey-martyp (approved):
Yeah looks like it just throws that in there as string literal if using the string separator param

PR #879, opened 2023-12-20
BSU rounds task form fix

kollil:
Amount units field is now visible.


brentlogan (approved):
Looks good for a short-term solution.

PR #897, opened 2024-01-14
Modified the process of determining cage size allowed.

Ohsudev:
Modified the process of determining cage size allowed.


kollil (approved):
Looks good

PR #903, opened 2024-01-17
Modified pregnancy date confirmtion on the Snapshot reports to not al…

Ohsudev:
Modified pregnancy date confirmtion on the Snapshot reports as not to allow displaying of pregnancy date when pregnancy was aborted.


kollil (approved):
Looks good

PR #906, opened 2024-01-17
Created Environmental Assessment table structures.

Ohsudev:
Created new Environmental Assessment table structures.


onprc_ehr/src/org/labkey/onprc_ehr/ONPRC_EHRModule.java
kollil, line 1: Raymond, Team city build has been failing since yesterday. I have an updated sql script version num # 23.008 in the code and the xml file is updated too. At some point, we both need to collaborate before we upload into prod.

kollil (commented):
Please see my comments.

PR #946, opened 2024-03-13
Modified SSRS Prime coding web link.

Ohsudev:
Modified SSRS server code to the new SSRS vm server.


brentlogan (commented):
I confirmed that primateapp3 is in the LabKey codebase 13 times and that each mention was replaced with pcdbssrsprd1.

kollil (approved):

PR #983, opened 2024-04-16
Housing transfers sorting

kollil:
Added sort order for the housing transfer data - area, room, cage

### DIFF
--- a/onprc_ehr/resources/queries/onprc_ehr/housing_transfers/.qview.xml
+++ b/onprc_ehr/resources/queries/onprc_ehr/housing_transfers/.qview.xml
@@ -1,0 +1,8 @@
+<customView xmlns="http://labkey.org/data/xml/queryCustomView" label="Housing Transfers Alert">
+    <sorts>
+        <sort column="building" />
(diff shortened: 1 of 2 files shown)


brentlogan (commented):
Lakshmi, wouldn't you want the sort to be from most general to most specific? I would have expected: area, building, room, cage. Maybe it doesn't matter in practice because each building has a unique name.

Also, does it make sense to either lock down the default view for the housing transfer alerts schema or use a locked-down named view for the link in the alert email? I was able to follow the link to the table, impersonate sidener, modify the sorting of the default view, and save it for everyone.

kollil:
@brentlogan, 
I think the sort order usually is Building ->area->room->cage. Building is not usually shown in many gird views, only area and rooms are seen. When I added building I put them in that order. 

Yes, You are right about the default view too. For now, I changed it on the UI in prod so users can have the sorted grid view until I install the update. I have the code written to lock the default view. Will deploy it today.

Thank you for the feedback.

Ohsudev (approved):

PR #984, opened 2024-04-18
Billing Period Summary / Discrepancy Report Automation

labkey-sweta:
#### Rationale
Automation test for https://docs.google.com/document/d/14OgIGqOSTUpQK9lyeebHrc_4i36cYDci-V7QLSlvM6o/edit

With all the update to data the report generated is 
image

Note: we stopped further test implementation upon client request before we could include a) rates to calculate the cost (hence, the Amount column is displaying 0.00) and b) Labwork results (hence, not displayed in the 'Charge summary' report under Category)

#### Related Pull Requests
* 

#### Changes
*


labkey-martyp (commented):
Looks good to me but would be good for @labkey-bpatel to check it out.

labkey-bpatel (approved):

PR #991, opened 2024-04-29
Birth trigger caching and misc fixes and test

labkey-martyp:
#### Rationale
Investigation of the demographic cache mismatches revealed a race condition with triggers processing births and refreshing the demographics cache. The affected area was the offspring inheriting certain records from the dam and caching that info in demographics cache. This PR reorders the birth trigger to avoid this situation.  Auto test added to validate the fix.

#### Related Pull Requests
* TBD

#### Changes
* Update order of birth trigger by not using ON_BECOME_PUBLIC event trigger.
* Add auto test validating the scenario
* Add additional test data
* Remove reference to Ext3 remove function
*


onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/AbstractGenericONPRC_EHRTest.java
```diff
@@ -287,6 +287,14 @@ protected void initProject() throws Exception
         cacheIds(Arrays.asList(MORE_ANIMAL_IDS));
     }
 
+//    @After
```
ankurjuneja, line 290: probably remove?

onprc_ehr/resources/data/divider_types.tsv
```diff
@@ -0,0 +1,10 @@
+divider	countAsSeparate	countAsPaired	countAsBoth	isMoveable	displaychar	short_description	bgcolor	border_style
+Mesh	TRUE	FALSE		TRUE	M	MESH	grey	dotted
+Solid	TRUE	FALSE		TRUE	S			solid
+Clear	TRUE	FALSE		TRUE	C	CLEAR		solid
+No Slide	TRUE	TRUE		TRUE	NO		yellow	none
+Pairing Divider	TRUE	FALSE		TRUE	P		grey	solid
+								
+								
+								
+
```
ankurjuneja, line 10: look like blank lines

ankurjuneja (approved):

PR #997, opened 2024-05-01
Update of Sheets for Public Rates and increase in F&A

jonesgaohsu:
#### Rationale
Update of Year Labels for years 1 thru 8  and update of FY&A Rate calculation
#### Related Pull Requests
* 

#### Changes
This is an annual event and each Public Rate Groups is updated in the XML


jonesgaohsu:
Naming was incorrect and needed to be updated as there was a hyphen instead of an underscore

PR #1043, opened 2024-06-06
Finance billing query update

kollil:
Removed the substitute paths for the billing containers to reference the onprc_billingPublic datasets and this change avoids user access to the Finance admin section.


jonesgaohsu (approved):
CHanges function as designed

PR #1066, opened 2024-07-11
Populate the Cln Obs

kollil:
Add vomit score to the cln observations


Ohsudev (commented):
I'm pretty sure we are not using this file.  If you trace the program from EHR admin site, it is use to populate ehr_lookups manually when we have to refresh Prime to a test database, but then I could be wrong.  Just check the coding scheme again for me.

Ben redesinged the funcitonality after he rewrote the program.

kollil:
Thanks for the feedback @Ohsudev, I have made the changes according to Ben's new process.

PR #1120, opened 2024-09-03
23.11 fb modify label gender to sex

Ohsudev:
Modified Prime's reporting label from Gender to Sex.


jonesgaohsu (approved):
Reviewed Code all looks as expected

PR #1124, opened 2024-09-06
Modified reporting program to show "gender" instead of "sex" as a var…

Ohsudev:
Modified reporting program to show "gender" instead of "sex" as a variable referencing data.  The changes will allow Necropsy Animal Details section to display the "Gender" value.


jonesgaohsu (approved):
Code looks functional

PR #1126, opened 2024-09-10
23.11 fb protocol update web part

jonesgaohsu:
#### Rationale
Building a Web part for eIACUC Process for the Integrity Container
#### Related Pull Requests
* 

#### Changes
Added 3 files to build the webpart eIACUCadmin


onprc_ehr/resources/schemas/onprc_ehr.xml
brentlogan, line 1: Gary, this file has both yours and Raymond's changes. It would be best if you reverted Raymond's changes in your codebase so they're only in Raymond's. That would eliminate the schema error message in your header.

PR #1153, opened 2024-10-08
Name Changes

kollil:
Changed the naming for the following:
 ARRS - > DCM
 Upper case to Mixed case of the titles:
 1. ART core
 2. ASB service request form
 3. PMIC service request form


brentlogan (approved):
1. I verified that you changed all instances of ARRS in the code to DCM.
2. I verified that you changed all all-caps link on the Enter Data screen.
Approved, and thanks!

brentlogan (commented):
Lakshmi, I happened to notice line 1324 of [ONPRC_EHRTest2.java](https://github.com/LabKey/onprcEHRModules/blob/23.11_fb_NameChanges/onprc_ehr/test/src/org/labkey/test/tests/onprc_ehr/ONPRC_EHRTest2.java). I don't know if the test code is case-sensitive, but if it is, you'll need to change this, too.

PR #1188, opened 2024-11-21
Modified Slide definitions reference image file

Ohsudev:
Modified Slide definitions reference image file


jonesgaohsu (approved):

jonesgaohsu:
All Data looks correct

PR #1193, opened 2024-11-26
Bigiron dev dependency

labkey-martyp:
#### Rationale
Pull in bigiron module jar for external devs not building premiumModules source.

#### Changes
* bigiron dependency


labkey-jeckels (approved):
We should presumably do the same for other EHR deployments on SQLServer, right?

labkey-martyp:
> We should presumably do the same for other EHR deployments on SQLServer, right?

Yes I'll do that. Just wanted to verify I wasn't missing anything here.

PR #1197, opened 2024-12-02
Process treatment_orders with a NULL enddate

brentlogan:
#### Rationale
Treatment orders with NULL end dates can be entered into PRIMe. 

`study.treatmentSchedule.sql` queries `study.treatment_orders` over a time range defined by "dependee" query `ehr_looksup.dateRange`. It doesn't return results of treatment orders where:

* `enddate` is NULL, or
* `enddate` is NULL AND a `dateRange` date is in the future.

`study.treatmentSchedule.sql` is used by printable report **Clinical Medications** and in the EHR animal history, **Daily Reports** > **Clinical Medications Schedule**. The animal history reports run for today's date (setting the start date of "dependee" query `ehr_looksup.dateRange` to today's date with a length of one). The printable report can be run for user-selected dates, including future dates and date ranges greater than one. 

`study.treatmentScheduleUpdate` is used for the TreatmentToDrug ETL, using today's date and range of 1 for the `dateRange`. 

#### Related Pull Requests
https://github.com/LabKey/onprcEHRModules/pull/1111

#### Changes
* The check for `dr.dateOnly` (all dates within dateRange) must be <= the treatment order's end date (if it exists).
* The check for `date` before `enddate` now returns TRUE if `enddate` is `NULL`.


jonesgaohsu (approved):
Code meets the objective and is written as was discussed

brentlogan:
We received feedback that the first commit didn't fully fix the issue. The printed report Clinical Medications failed to produce the desired results for treatment orders having no end date when the report's date was in the future. I reviewed `study.treatmentSchedule` and realized this results in no overlap between `t1.date` to `t1.enddateCoalesced` and the range defined by `dr` (which is the future date with a range of 1 day), and thus, the treatment order doesn't show in the report.

This latest commit modifies only `study.treatmentSchedule` and not `study.treatmentScheduleUpdate`, which might need to be updated.

brentlogan:
@bbimber please take a look at this pull request. I've determined that allowing treatment orders to have NULL end dates breaks the two queries currently in the pull request. I believe these two proposed changes won't impact the rest of PRIMe, but I'd like your feedback. It might make more sense to require end dates for all treatment orders to avoid this issue.  

Should you agree that these changes are fine, I propose making similar changes to the other `study.treatmentScheduleX` queries that are only slight modifications of `treatmentSchedule`. I'd also add the second change to `study.treatmentScheduleUpdate` (which has only one change in this pull request in its current form).

Thanks!

onprc_ehr/resources/queries/study/treatmentSchedule.sql
```diff
@@ -84,8 +84,9 @@ SELECT
 FROM ehr_lookups.dateRange dr
 
 JOIN study."Treatment Orders" t1
-  --NOTE: should the enddate consider date/time?
-  ON (dr.dateOnly >= t1.dateOnly and dr.dateOnly <= t1.enddateCoalesced AND
+  --COALESCE returns results for treatment orders without an enddate and with s.date in the future. This is needed for
+  --printable report Clinical Medications
+  ON (dr.dateOnly >= t1.dateOnly AND dr.dateOnly <= COALESCE(CAST(t1.enddate AS DATE), dr.dateOnly) AND
```
bbimber: This is based on taste and what I think is readability, but I would do an explicit test for NULL rather than COALESCE() it. The intention behind COALESCE() doesnt seem as clear to me, if someone reads this in the future. so something like:
```
ON (dr.dateOnly >= t1.dateOnly AND (dr.dateOnly <= t1.enddate OR t1.enddate IS NULL))
```
but the existing code isnt wrong per se. this suggestion makes it more in-line with your change below

bbimber (approved):
one minor comment, but i think this will work. I'm approving since the suggestion isnt required for the code to work

onprc_ehr/resources/queries/study/treatmentSchedule.sql
brentlogan: @bbimber, thanks for reviewing the code. I prefer your suggestion as it's cleaner and self-documenting the issue.

jonesgaohsu (approved):
Reviewed changes with Brent and this is ready for production

PR #1200, opened 2024-12-05
Fix URL in MHC dashboard

bbimber:


jonesgaohsu (approved):
Per request from Ben, this has no direct effect on PRIMe

PR #1207, opened 2024-12-14
Created scripting program to update compliance string "ARRS " to "DCM"

Ohsudev:
Convert all compliance character name strings from "arrs" to "dcm".


jonesgaohsu (approved):
Fils appear correct

PR #1226, opened 2025-01-15
Migrate <permissionClasses> -> <requiresPermissions>

labkey-adam:
#### Rationale
We recommend `<requiresPermissions>` since it matches the other elements and action annotations; use what we document so we propagate the recommended element. `<permissionsClasses>` remains a valid synonym.

### DIFF
--- a/onprc_ehr/resources/views/ehrAdmin.view.xml
+++ b/onprc_ehr/resources/views/ehrAdmin.view.xml
@@ -1,7 +1,7 @@
 <view xmlns="http://labkey.org/data/xml/view" title="EHR Administration">
-    <permissionClasses>
+    <requiresPermissions>
         <permissionClass name="org.labkey.api.security.permissions.AdminPermission"/>
-    </permissionClasses>
+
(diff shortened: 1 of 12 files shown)


onprc_ehr/resources/views/ehrAdmin.view.xml
```diff
@@ -1,7 +1,7 @@
 <view xmlns="http://labkey.org/data/xml/view" title="EHR Administration">
-    <permissionClasses>
+    <requiresPermissions>
         <permissionClass name="org.labkey.api.security.permissions.AdminPermission"/>
-    </permissionClasses>
+
```
labkey-martyp: Closing tag?

labkey-martyp (approved):
Missing one closing tag

onprc_ehr/resources/views/ehrAdmin.view.xml
labkey-adam: Uh, good point!

PR #1267, opened 2025-02-26
SLA purchase protocol fix

kollil:
Removed the breeding info from the protocol selection drop down list.


brentlogan (commented):
Approved.

brentlogan (approved):

PR #1272, opened 2025-03-03
Remove CSS classes from Vet Review

labkey-martyp:
#### Rationale
CSS classes were added in an inline scripting refactor. These classes are on the core EHR vet review but that styling is not wanted here.

#### Changes
* Remove CSS classes


ankurjuneja (dismissed):

jryurkanin:
@labkey-martyp @ankurjuneja  Was it necessary to also get rid of the vrdc-row class? Brent noticed some undesired effects of this fb as it currently is. It removes the behavior of opening a Clinical Remarks pop-up, and Brent thinks it's related to removing the vrdc-row class here.

labkey-martyp:
> @labkey-martyp @ankurjuneja Was it necessary to also get rid of the vrdc-row class? Brent noticed some undesired effects of this fb as it currently is. It removes the behavior of opening a Clinical Remarks pop-up, and Brent thinks it's related to removing the vrdc-row class here.

Apologies, I just removed the css classes that were added, I didn't see that the vrdc-row was being used for the handler. I added it back.

PR #1325, opened 2025-04-16
Remove code that was originally set up for "Breeder Lock Release Grou…

Ohsudev:
Modified program to remove code that allowed Admin permission group to remove a "lock" onto Birth screen when
a current user has exclusive access to the birth screen.  The permission Group is Called "Birth Lock Release Group".


Ohsudev:
converted to 25.11 version

PR #1353, opened 2025-05-13
24.11 fb tissue distribution historical updates

Ohsudev:
Created a  new program that will update currently recorded Tissue Distribution records with task ids so that the
records can be edited by Pathology staff.


kollil (dismissed):

kollil (dismissed):

jonesgaohsu (approved):
Code reviewed and no issues found

PR #1403, opened 2025-07-27
Skip inserting data when creating empty schemas - EHR-related tables

labkey-adam:
#### Related Pull Requests
* https://github.com/LabKey/platform/pull/6867

### DIFF
--- a/mergesync/resources/schemas/dbscripts/postgresql/mergesync-0.05-0.06.sql
+++ b/mergesync/resources/schemas/dbscripts/postgresql/mergesync-0.05-0.06.sql
@@ -29,6 +29,7 @@ CREATE TABLE mergesync.mergetolkmapping (
   CONSTRAINT PK_mergetolkmapping PRIMARY KEY (mergetestname)
 );
 
+-- @SkipOnEmptySchemasBegin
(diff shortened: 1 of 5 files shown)


mergesync/resources/schemas/dbscripts/postgresql/mergesync-0.05-0.06.sql
```diff
@@ -77,7 +78,6 @@ INSERT INTO mergesync.testnamemapping (servicename, mergetestname, automaticresu
 INSERT INTO mergesync.testnamemapping (servicename, mergetestname, automaticresults) VALUES ('Urinalysis', 'URINE', false);
 INSERT INTO mergesync.testnamemapping (servicename, mergetestname, automaticresults) VALUES ('Yersinia culture', 'YERSNA', false);
 
-TRUNCATE TABLE mergesync.mergetolkmapping;
```
labkey-jeckels, line 80: I'm not understanding why this line is getting deleted. In the normal bootstrap scenario, mergesync-0.04-0.05.sql will have inserted rows.

labkey-adam: The table was dropped and recreated just before this INSERT chunk, so the TRUNCATE is a no-op. Not causing any harm, of course.

labkey-jeckels (approved):

PR #1422, opened 2025-08-16
25.7 fb pathology scheduler amended

Ohsudev:
Merging previous code from 24.11 to 25.7.  Project is still in a testing state.


Ohsudev:
converted to 25.11 version

PR #1441, opened 2025-08-30
Modified Compliance primary page so that navigation selection menu it…

Ohsudev:
Modified program to allow navigation menu link to open a new tab browser screen.


Ohsudev:
Converted from 24.11 to 25.7 coding.

PR #1455, opened 2025-09-06
25.7 fb compliance completiondate updates

Ohsudev:
Created a new program that will update the container value from entries made to the Compliance Access folder so that the
resulting entries gets uploaded to the Compliance main page.


Ohsudev:
No longer needed

PR #1457, opened 2025-09-06
25.7 fb compliance open browser

Ohsudev:
Created a new program allow navigational menus to open a new tab browser when selected.


Ohsudev:
No longer needed.

PR #1465, opened 2025-09-11
Archive audit log

brentlogan:
#### Related Pull Requests
* 

#### Changes
*


brentlogan:
@jonesgaohsu I discovered onprc_ehr-25.000-25.001.sql still had both the CREATE and ALTER portions of code, even though both of us thought they had been deleted. Our changes yesterday hadn't removed the primary key from that portion of code, and that's what PRIMe was complaining about when you installed this morning. I reverted your changes and removed the ALTER section. 

Additionally, I'm going to rename the destination database on pcdbdev1 so we can use the same stored procedure everywhere.

jonesgaohsu (approved):
All changes are good and this has been well tested

PR #1482, opened 2025-09-26
Fix treatment order updates

labkey-martyp:
#### Rationale
Updating treatment orders is failing on duplicate ParticipantSequenceNum. We shouldn't be passing that in anyway. Also need better error handling.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/7076

#### Changes
* Remove ParticipantSequenceNum
* Check for null return and errors


labkey-jeckels (approved):
I didn't see any repro steps so I didn't attempt testing. Let me know if you'd like some.

labkey-martyp:
> I didn't see any repro steps so I didn't attempt testing. Let me know if you'd like some.

User tested. Thanks.

PR #1484, opened 2025-09-29
Fix ABV process notification

brentlogan:
#### Rationale
The current ABV process notification generates false positives because MMA's ABV calculations go into PRIMe right before the hour and the way the notification determines if the calculations are stale.


jonesgaohsu (approved):
Reviewed with Brent

PR #1492, opened 2025-10-05
Fix stale ABV notification

brentlogan:
#### Rationale
The last update didn't use the correct compare type, resulting in continued false positives.

#### Related Pull Requests
* 

#### Changes
*


jonesgaohsu (approved):
Reviewed code changes with BRent No Issues

PR #1499, opened 2025-10-09
Modified program to include "transactionNUmber" as part of the

Ohsudev:
invoice items "Reverse/Adjustments editing process to include "transaction number" values. Currently, still testing stage.


Ohsudev:
No longer needed

PR #1502, opened 2025-10-13
Refactor per diem queries for perf

labkey-martyp:
#### Rationale
The per diem queries using perdiemsByDay were slow and getting much slower as data grew. This PR does some refactoring to do less per-row processing. Also some 0 checks added for divide operations.

#### Changes
* Factor complex operations out of CASE WHEN conditions
* Move joins out of subqueries
* Add some NULLIFs


jonesgaohsu:
Reviewed the code and loaded on test server with great results.

jonesgaohsu (approved):
Reviewed and also loaded to test instance and saw marked performance improvement

PR #1213, opened 2024-12-20
23.11 fb eiacuc to prime end date

jonesgaohsu:
#### Rationale
We want to use eIACUC2 data (the source of truth for protocols) to keep `ehr.protocol` automatically updated. We see at least three steps in this process:

- Ensure each base protocol family in `ehr.protocol` has a non-NULL `enddate` where the most recent ever-approved protocol per base protocol family is not currently approved.
    - `ehr.protocol` has only one record per base protocol family, using the base protocol's protocol_ID and the most recently approved protocol's other data.
    - most recent = latest `approval_date`
    - ever approved = `PROTOCOL_state` is one of approved, terminated, or expired 
    - The end date indicates that no protocol in the base protocol family is *currently* approved. It doesn't necessarily indicate when that first occurred.
- Update `ehr.protocol` when eIACUC2 has a more recent approved protocol for a base protocol family.
- Create a new base protocol record in `ehr.protocol` when eIACUC2 data has a new protocol that is not a renewal.

This pull request addresses only the first step.

#### Related Pull Requests
* 

#### Changes
* Add `BaseProtocol`, `RevisionNumber`, and `NewestRecord` fields to existing table `onpre_ehr.eIACUC_PRIME_VIEW_PROTOCOLS`. 
* Create a stored procedure that adds end date information to selected records in `ehr.protocol`.

PR #1510, opened 2025-10-15
Simplify eIACUC Direct Connect queries

brentlogan:
#### Rationale
The ETL loading eIACUC data into the EHR now determines the base protocol and revision number. These queries were repeating that work. 

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/1213

#### Changes
* Use **Base Protocol** and **Revision Number** as calculated by the ETL
* Delete unnecessary query
* Join `ehr.protocol` instead of `ehr.project`


jonesgaohsu (approved):
Review this with Brent and all is good

PR #1515, opened 2025-10-21
Adopt FileLike for more pipeline APIs

labkey-jeckels:
#### Rationale
Adopt FileLike for more pipeline APIs

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/7134

#### Changes
* Many switches to `FileLike`


onprc_billing/src/org/labkey/onprc_billing/pipeline/BillingTask.java
```diff
@@ -467,19 +403,14 @@ private List<Map<String, Object>> getRowList(Container c, String schemaName, Str
         ts.setNamedParameters(params);
 
         final List<Map<String, Object>> rows = new ArrayList<>();
-        ts.forEach(new Selector.ForEachBlock<>()
-        {
-            @Override
-            public void exec(ResultSet object) throws SQLException
+        ts.forEach(object -> {
+            Results rs = new ResultsImpl(object, colKeys);
```
labkey-matthewb: This is convoluted, we can just use ColumnInfo.getValue(ResultSet) 

ts.forEach(rs -> {
            Map<String, Object> ret = new HashMap<>();
            for (FieldKey fk : colKeys.keySet())
            {
                ret.put(fk.toString(), fk.getValue(rs);
            }

labkey-matthewb (changes requested):

onprc_billing/src/org/labkey/onprc_billing/pipeline/BillingTask.java
labkey-jeckels: That's what I get for accepting an auto-refactor -> lamba. I updated this old code.

labkey-matthewb (approved):

PR #1521, opened 2025-10-23
25.7 fb clinpath panic notification

Ohsudev:
New notification program to flag chemistry values that contains the text string called "Panic values" as part of lab
results.


Ohsudev:
converted to 25.11 version

PR #1526, opened 2025-10-28
25.7 fb parentage evaluation

Ohsudev:
Modified the program that determines the dam/sire values that populates the "Parent Information" from the
Animal Snapshot reports.

### DIFF
--- a/onprc_ehr/resources/queries/study/parentageSummary.sql
+++ b/onprc_ehr/resources/queries/study/parentageSummary.sql
@@ -21,7 +21,7 @@ SELECT
   p.method
 
 FROM study.parentage p
-WHERE p.qcstate.publicdata = true and p.enddateCoalesced <= now()
+WHERE p.qcstate.publicdata = true and p.enddate is null
 
 UNION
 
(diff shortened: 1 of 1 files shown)


kollil (dismissed):

kollil (approved):

bbimber:
@brentlogan, @Ohsudev, and/or @kollil: I dont quite understand these changes. Specifically:

1) It is almost always better practice to use the "coalesced" version of enddates, unless you have a very specific reason. The purpose of testing "p.enddateCoalesced <= now()", instead of what you changed it to be (enddate IS NULL), is that the former would correctly handle a future end dates. While I will grant that this situation is not likely in the case of parentage; however, there is value in consistency and in writing code that is robust to unexpected conditions.  

2) The original intend of this query was to report all parentage data. That means that if a given animal was reported with one dam, but genetic data contradicted this, they are nonetheless both returned. Other queries in EHR, demographicsParents.sql for example, coalesce between the sources of parentage and preferentially take genetic over observed parentage. The primary piece of code that seems to use parentageSummary is ParentsDemographicsProvider. What's driving the need for this?

PR #1527, opened 2025-10-28
Add row count validation to ETL

bbimber:
This adds an addition check to the existing MHC ETL. After the ETL it compares row counts on the source/destination. I find this to be a useful check to alert if an ETL isnt behaving as expected.


brentlogan (approved):

bbimber:
Thanks @brentlogan. This check may or may not be useful to consider for other ETLs.

PR #1528, opened 2025-10-29
Update archive audit log stored procedure

brentlogan:
#### Rationale
The stored procedure in the related PR didn't create the audit schema correctly. This PR corrects that.

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/1465

#### Changes
* Modify stored procedure ArchiveAuditLogs.
* Rev ONPRC_EHR module's schema.

PR #1536, opened 2025-11-09
Add @RetentionYears parameter to ArchiveAuditTables stored procedure

brentlogan:
#### Rationale
The ArchiveAuditTables stored procedure effectively prevents PRIMe from being used when it runs by blocking writes to the audit log when a table is being archived. 

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/1465
* https://github.com/LabKey/onprcEHRModules/pull/1528

#### Changes
* Create a @RetentionMonths parameter and archive data older than the value. 
* Decrement @RetentionYears each run, starting at 144 and stopping at 12.

### DIFF
--- a/onprc_ehr/resources/etls/HarvestToPrime_Process.xml
+++ b/onprc_ehr/resources/etls/HarvestToPrime_Process.xml
@@ -71,7 +71,7 @@
 
     <schedule>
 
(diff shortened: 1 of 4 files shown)


jonesgaohsu (approved):
Stored Proceded was verified and functions as designed

PR #1573, opened 2025-12-11
Modified scheduler program for Radiology so that the comment informat…

Ohsudev:
Modified Radiology Scheduler to include comment information as part of the clendar interface.


Ohsudev:
converted to 25.11 version

PR #1578, opened 2025-12-16
Update from Original Fix by Raymond for Source display in invoiced Items

jonesgaohsu:
#### Rationale
Based on Work completed by Raymond to display correct value for Source Record and update to Reverse Charge 
#### Related Pull Requests
* 

#### Changes
onprc_billilng.xml
qview.xml
ReverseChargeWindow.js


brentlogan:
@jonesgaohsu, I believe you can close this PR. I already deployed a FB that was based on this.

PR #1589, opened 2025-12-24
25.7 fb surgical rounds observation updates

Ohsudev:
Modified Surgical Rounds Clinical Observation section to allow wrapping text feature when entering data.


Ohsudev:
converted to 25.11 version.

PR #1599, opened 2026-01-08
Update to ChargeRate.js and active view xml to have end time at the …

jonesgaohsu:
…end of the day

#### Rationale
Update to change the logic so the end date of a charge item is the end of the day.  Also changed the logic for active charges to use coaleasced

#### Related Pull Requests
* 

#### 
ChangeschargeRate.js
active Rate.qview.xml


brentlogan (approved):
Looks good.

PR #1628, opened 2026-02-04
Colony management alert update

kollil:
Display any mismatched observed and genetic Dams


brentlogan (approved):
I think there might be a small edge case when an offspring has two genetic dams, but this is good to deploy now, and fix the edge case if it's important to Cassandra.

PR #1636, opened 2026-02-13
Make the search more specific

labkey-klum:
#### Rationale
The test was matching more search results than intended, change the search term so just the 2 issues are returned.


labkey-klum:
@labkey-danield @labkey-tchad Let me know if you think this needs to target a different version.

labkey-tchad:
I'd like this fix in 25.11. That version is going to be active for PRCs for a while and it is [intermittently failing there](https://teamcity.labkey.org/test/2913470746468526406?currentProjectId=LabKey_2511Release) for the same reason.

PR #1643, opened 2026-02-24
Parentage dam mismatch in Colony alert

kollil:
Added Parentage dam mismatch report in colony alert.


brentlogan (approved):
I'm approving this PR with the understanding that we'll confer with Raymond about the assumed rare edge case of ignored multiple genetic dams caused by the `MAX` aggregate functions in lines 26 and 37 in `ParentageDamMismatch.sql`.

PR #1656, opened 2026-03-16
Room utilization past dates

brentlogan:
#### Rationale
PRIMe's standard room utilization query runs only with current data. This provides similar data, but for past dates. As this query needs to provide data prior to the start of onprc_ehr.CageAuditLog, this query cannot provide historic room configuration information. It uses current information to count cage locations per room.

#### Related Pull Requests
* 

#### Changes
*


brentlogan:
Originally requested by Lisa. No mo' need. :-(

PR #1676, opened 2026-03-31
25.11 fb radiology scheduler events

Ohsudev:
Modified Radiology scheduler program so that the comments values are viewable within the calendar
panel.


Ohsudev:
Converted to Prime 26.3 coding schemes

PR #1677, opened 2026-03-31
25.11 fb surgical rounds observation update

Ohsudev:
Modified Surgical Rounds Observation panel to allow editing as a text area controls.


Ohsudev:
Converted Prime 26.3 coding schemes

PR #1707, opened 2026-05-01
Disable Id on Exam forms when case created.

labkey-martyp:
#### Rationale
[Ticket 54568](https://www.labkey.org/ONPRC/Support%20Tickets/issues-details.view?issueId=54568): Exams/Cases allows creation of observation linked to another animal's case.

This PR disables the Animal Id data entry field in the SOAP (clinremarks) section of Exams/Cases and BSU Exam when a case is created. When a case is created the caseid is attached to all records in the form. The only way to change the Id in the forms (which could cause multiple Ids on the same case) is via the SOAP section Id, so disabling it when a case is associated with the records ensures no other Ids can be entered for the case.

#### Changes
* Disable the Id field if there is already a caseid in the record
* Disable the Id field when the 'casecreated' event is fired
* Remove duplicate remark metadata
* Remove CagemateClinicalReportFormType as it is unused
* Create automated tests


onprc_ehr/resources/web/onprc_ehr/model/sources/CaseMgmt.js
```diff
@@ -0,0 +1,57 @@
...
+                                var el = field.getEl();
+                                if (el){
+                                    el.set({'data-qtip': readOnly ? TOOLTIP : ''});
+                                }
+                            };
+                            var syncReadOnly = function(){
+                                var rec = EHR.DataEntryUtils.getBoundRecord(field);
+                                var readOnly = !!(rec && rec.get('caseid'));
+                                field.setReadOnly(readOnly);
+                                setTooltip(readOnly);
```
labkey-bpatel, line 46: Does it make sense to gray out the disabled Id? so that it doesn't look like it is available for updating - just a little better visually/user experience, imo.

labkey-bpatel (dismissed):
Looks good, please review my comment.

onprc_ehr/resources/web/onprc_ehr/model/sources/CaseMgmt.js
labkey-martyp: yes done thanks

labkey-bpatel (approved):

labkey-martyp (commented):
Client approved to merge in.

PR #1712, opened 2026-05-08
Modified Cage layout since the monkey id is showing as being too smal…

Ohsudev:
Modified Cage Room Layout web page in order to enlarged font size of the monkey id being displayed on the form.


Ohsudev:
Converted the same code to Prime 26.3

PR #1759, opened 2026-06-13
Scope ONPRC billing run deletion to the current container

labkey-martyp:
#### Rationale

ONPRC_BillingManager.deleteBillingRuns built SimpleFilters over the raw onprc_billing tables (invoicedItems, invoiceRuns, miscCharges) from request-supplied primary keys with no container predicate, and the method received no Container. Since DeleteBillingPeriodAction's permission check is evaluated only against the current folder, a billing admin in one folder could delete or detach billing rows belonging to any other folder by POSTing arbitrary object IDs — a cross-container horizontal privilege-escalation bug. This change confines every operation to the caller's folder.

#### Related Pull Requests

* _none_

#### Changes

* Add a Container parameter to deleteBillingRuns and pass getContainer() from both DeleteBillingPeriodAction call sites.
* Add a container filter to every read, delete, and update against invoicedItems, invoiceRuns, and miscCharges (including the Table.update that detaches misc charges from a deleted invoice).
* Add integration test ONPRC_BillingManager.TestCase asserting that a delete issued from one container cannot affect another, and register it in ONPRC_BillingModule.getIntegrationTests().
* Drop two unused row-count locals from the Table.delete calls.

### DIFF
--- a/onprc_billing/src/org/labkey/onprc_billing/ONPRC_BillingController.java
+++ b/onprc_billing/src/org/labkey/onprc_billing/ONPRC_BillingController.java
@@ -162,7 +162,7 @@ public class ONPRC_BillingController extends SpringActionController
             Set<String> ids = DataRegionSelection.getSelected(form.getViewContext(), true);
 
             StringBuilder msg = new StringBuilder("You have selected " + ids.size() + " billing runs to delete.  This will also delete: <p>");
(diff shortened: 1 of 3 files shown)


labkey-bpatel (dismissed):

labkey-martyp:
Closing in favor of #1767.

PR #1768, opened 2026-06-18
Created a new Animal History report tab called

Ohsudev:
Created a new Animal History tab reports that shows Clinical Encounters "TB Test Intradermal" entries 
from both Procedure, and Surgical entries.


ohsudev-jonesga (approved):
No issue with Coded

PR #1777, opened 2026-06-22
Modified Bulk Clinical Entry to contain Pairing Observation Entry form.

Ohsudev:
Created a new Pairing Observation as an additional input form to the "Bulk Clinical Entry" input form set.


ohsudev-jonesga (approved):
Review completed no issues seen

PR #1778, opened 2026-06-22
Fix stored XSS in finance notification charge summary report

labkey-martyp:
## Rationale

FinanceNotification.writeResultTable and its overriding subclass DCMFinanceNotification.writeResultTable are divergent copies of the ehr_billing BillingNotification report that concatenated editor-entered database values (investigator, project, alias/account, OGA project number, category) and their derived URLs directly into HTML with no escaping. That HTML is rendered verbatim into the LDK RunNotificationAction admin preview via HtmlString.unsafe and is also sent as the HTML email body, so a stored payload in any of those project/alias fields executed in the browser of any user who previewed the notification or received the email — a stored XSS with privilege-escalation potential toward admins. Both are registered, active notifications (ONPRC_BillingModule). This is the ONPRC counterpart to the BillingNotification fix; unlike that copy, the top category summary table here was also unescaped.

## Related Pull Requests

- LabKey/ehrModules#1159

## Changes

- FinanceNotification: wrap every editor-entered value (investigator, project, account, project number, category) and its derived href URL in PageFlowUtil.filter in the per-financial-analyst tables, and filter the top category summary table (url and category), which was unescaped in this copy.
- DCMFinanceNotification: apply the same filtering to its writeResultTable override, which the parent-class fix did not cover; mapped to this override's token layout (tokens[0] project, tokens[1] alias, tokens[2] OGA project number).
- Add the org.labkey.api.util.PageFlowUtil import to both files.

### DIFF
--- a/onprc_billing/src/org/labkey/onprc_billing/notification/DCMFinanceNotification.java
+++ b/onprc_billing/src/org/labkey/onprc_billing/notification/DCMFinanceNotification.java
@@ -2,6 +2,7 @@ package org.labkey.onprc_billing.notification;
 
 import org.apache.commons.lang3.StringUtils;
 import org.labkey.api.data.Container;
+import org.labkey.api.util.PageFlowUtil;
 import org.labkey.onprc_billing.ONPRC_BillingManager;
 import org.labkey.onprc_billing.ONPRC_BillingSchema;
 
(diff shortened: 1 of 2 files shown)


labkey-martyp:
Closing in favor of #1779, which applies the same fix based on release26.3-SNAPSHOT.

PR #1780, opened 2026-06-22
Modified Clinical Medications Online version of the Printable web sit…

Ohsudev:
Modified the Clinical Medication online version of the Printable reports so that information generated by
pdf mirrors the same information when online version is selected.


ohsudev-jonesga (dismissed):
Update appears in order

kollil (approved):

PR #1792, opened 2026-07-03
Preserve delete audit records for MHC data

bbimber:
@brentlogan: this PR is designed to help the ETL that syncs MHC data from prime-seq to prime. there is currently a loophole that lets some source record deletes get missed. this should fix that. it will be deployed to prime-seq, so doesnt need anything from PRIME.


brentlogan (approved):
Looks good

PR #1833, opened 2026-08-04
Created a new stored procedure program that will process all TB Test …

Ohsudev:
Create a new stored procedure that creates new Clinical Observation TB Test entries from entries made from
"TB Test Intradermal" historical records.


Ohsudev:
No longer needed.  Too many issues with this PR.  Will create a new PR.

